PR #29, opened 2019-10-22
DCA-206:Fix create issue with one issue type

mmizin:


app/selenium_ui/jira/modules.py
```diff
@@ -127,10 +127,11 @@ def __filer_epic(element):
 
             webdriver.find_element_by_id(ISSUE_TYPE_DROPDOWN).click()
             issue_elements = webdriver.find_elements_by_class_name("aui-list-item")
-            filtered_issue_elements = list(filter(__filer_epic, issue_elements))
-            rnd_issue_type_el = random.choice(filtered_issue_elements)
-            action = ActionChains(webdriver)
-            action.move_to_element(rnd_issue_type_el).click(rnd_issue_type_el).perform()
+            if issue_elements:
```
ometelytsia: When issue_elements has only one element, what would be a logic?

mmizin: We just move ahead. We don't need to choose an issue element because it already chose as default.

ometelytsia: if issue_elements - will return True, so logic will not move ahead.

mmizin: if issue_elements won't be empty it means that we have more than one issue element. Issue elements don't include element that we see without clicking on drop-down

jkim2-atlassian: If I understand correctly, nothing will be returned (no aui-list-item), if we only have one issue type, right? So we just proceed without clicking on the dropdowns.

I think it would be better if we rename issue_elements to issue_elements_in_drodown to avoid confusion.

ometelytsia: I like Jun's suggestion. With current naming, it is not clear that `issue_elements` returns None in case of only one element.

mmizin: That's so true! I will rename it as Jun suggested

PR #31, opened 2019-10-23
DCA-110: improve reading yml config file

mmizin:


app/util/conf.py
```diff
@@ -0,0 +1,38 @@
...
+        obj = read_yml_file(JIRA_YML)
+        self.hostname = obj['settings']['env']['application_hostname']
+        self.protocol = obj['settings']['env']['application_protocol']
+        self.port = obj['settings']['env']['application_port']
+        self.postfix = obj['settings']['env']['application_postfix'] or ""
+        self.admin_login = obj['settings']['env']['admin_login']
+        self.admin_password = obj['settings']['env']['admin_password']
+        self.concurrency = obj['settings']['env']['concurrency']
+
+    def __repr__(self):
```
ometelytsia: why do we need this method?

app/util/conf.py
```diff
@@ -0,0 +1,38 @@
...
+        return f'hostname = {self.hostname!r}\n' \
+               f'protocol = {self.protocol!r}\n' \
+               f'port = {self.port!r}\n' \
+               f'postfix = {self.postfix!r}\n' \
+               f'admin_login = {self.admin_login!r}\n' \
+               f'admin_password = {self.admin_password!r}\n' \
+               f'concurrency = {self.concurrency!r}\n'
+
+    def get_server_url(self):
```
ometelytsia: what do you think about converting this method to a property?

app/util/conf.py
```diff
@@ -0,0 +1,38 @@
...
+        return f'{self.protocol}://' \
```
ometelytsia: can we have a one-liner here

app/util/conf.py
mmizin: It's like a best practice. It can be also useful for debugging.

app/util/conf.py
mmizin: Yes, it's a good idea !!! Thanks for the suggestion )

app/util/conf.py
mmizin: Yes, of course, I've done it for the best readability. But can change it!

mmizin: Now it looks even better !!!

app/selenium_ui/conftest.py
```diff
@@ -46,17 +47,6 @@
         pass
 
 
-def application_url():
-    with JIRA_YML.open(mode='r') as fs:
-        jira_yaml = yaml.load(fs, Loader=yaml.FullLoader)
```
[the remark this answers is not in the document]
mmizin: done!

app/selenium_ui/jira/modules.py
```diff
@@ -10,12 +10,13 @@
 from selenium.webdriver.support.ui import Select
```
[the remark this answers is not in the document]
mmizin: done!

app/selenium_ui/conftest.py
```diff
@@ -1,6 +1,7 @@
 import atexit
 import csv
 import datetime
+import glob
```
ometelytsia: why do you need this package?

app/selenium_ui/jira/modules.py
```diff
@@ -130,7 +128,7 @@ def __filer_epic(element):
             webdriver.find_element_by_id(ISSUE_TYPE_DROPDOWN).click()
             issue_elements_in_dropdown = webdriver.find_elements_by_class_name("aui-list-item")
             if issue_elements_in_dropdown:
-                filtered_issue_elements = list(filter(__filer_epic, issue_elements))
+                filtered_issue_elements = list(filter(__filer_epic, issue_elements_in_dropdown))
```
ometelytsia: I believe this was fixed in another PR.
Just merge with dev to get this fix, but not make it here.

app/selenium_ui/conftest.py
mmizin: I thought it was before.
I will remove it

app/util/data_preparation/confluence/prepare-data.py
```diff
@@ -39,13 +33,13 @@ def write_to_file(file_name, list):
 def main():
     print("Started preparing data")
 
-    url = sys.argv[1]
+    url = CONFLUENCE_SETTINGS.server_url
```
ometelytsia: If we are going with this approach.
1. We need to change prepare_data.py call in jira.yml:
from
python util/data_preparation/jira/prepare-data.py `${application_protocol}://${application_hostname}:${application_port}${application_postfix}`
to
`python util/data_preparation/jira/prepare-data.py`
2. Prepate date should fail gracefully if jira.yml does not have correct values for host, port, ect. With explanation of what and where has to be set.

app/selenium_ui/jira/modules.py
mmizin: I found out that I've not changed this variable.

PR #36, opened 2019-10-31
DCA-216: User-friendly message for CAPTCHA

astashys:


app/util/data_preparation/jira/api.py
```diff
@@ -21,6 +21,19 @@ def __init__(self, host, user, password, api_session=None, timeout=30):
...
+    @staticmethod
+    def __verify_response(response, error_msg):
+        if response.ok:
+            return
+
+        status_code = response.status_code
+        if status_code == 403:
+            denied_reason: str = response.headers.get('X-Authentication-Denied-Reason')
+            if denied_reason and denied_reason.startswith('CAPTCHA_CHALLENGE'):
+                raise Exception("The user name is in Captcha Mode. Please login via Web UI first and re-run tests.")
```
ometelytsia: Nitpick: Maybe, it would be useful to add an exact username into the message, to make a manual login easier.

astashys: it makes sense

PR #37, opened 2019-11-04
DCA-210: Tech Debt: use OOP approach for ApiJira and ApiConfluence

astashys:


astashys:
It seems like we can remove folders util/data_preparation/jira and util/data_preparation/confluence and just rename prepare-data.py files to jira_prepare_data nad confluence_prepare_data accordingly.

PR #46, opened 2019-11-13
DCA-240:Refactor Full name creation of confluence Users

mmizin:


app/util/data_preparation/api/confluence_clients.py
```diff
@@ -100,10 +90,9 @@ def create_user(self, username=None, password=None):
         token = proxy.confluence2.login(self.user, self.password)
 
         if not proxy.confluence2.hasUser(token, username):
-            names = self.__get_random_names(2)
             user_definition = {
                 "email": f"{username}@test.com",
-                "fullname": f"{names[0]} {names[1]}",
+                "fullname": f"{username}_full_name",
```
ometelytsia: I think `"fullname": username.capitalize()` would have more sence.

mmizin: As u wish

PR #51, opened 2019-11-20
None/check indexsnapshot size

SergeyMoroz0703:
Added verification of IndexSnapshot minimal size for index-snapshot.sh

### DIFF
--- a/app/util/confluence/index-snapshot.sh
+++ b/app/util/confluence/index-snapshot.sh
@@ -17,7 +17,7 @@ FAIL_FAST_ATTEMPTS=15
 
 
 while [ ${COUNTER} -lt ${ATTEMPTS} ];do
-    if sudo su -c "test -f ${SNAPSHOT}"; then
+    if sudo su -c "test -f ${SNAPSHOT}" && find ${SNAPSHOT} -type f -size +5G 2>/dev/null | grep -q .; then
         echo # New line
         echo "Snapshot was created successfully."
         break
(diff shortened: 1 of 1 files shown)


app/util/confluence/index-snapshot.sh
```diff
@@ -17,7 +17,7 @@ FAIL_FAST_ATTEMPTS=15
 
 
 while [ ${COUNTER} -lt ${ATTEMPTS} ];do
-    if sudo su -c "test -f ${SNAPSHOT}"; then
+    if sudo su -c "test -f ${SNAPSHOT} && find ${SNAPSHOT} -type f -size +${MIN_SNAPSHOT_SIZE} 2>/dev/null | grep -q ."; then
```
ometelytsia: Maybe it would be better to have simpler logic with getting size of file and if statement?

app/util/confluence/index-snapshot.sh
```diff
@@ -17,7 +17,8 @@ FAIL_FAST_ATTEMPTS=15
 
 
 while [ ${COUNTER} -lt ${ATTEMPTS} ];do
-    if sudo su -c "test -f ${SNAPSHOT}"; then
+    size=`sudo su -c "ls -la ${SNAPSHOT}" | awk '{print $5}'`
```
ometelytsia: Let's use the same approach for getting file size
SNAPSHOT_SIZE=`sudo su -c "du -s ${SNAPSHOT}" | cut -f1"` - size in KB
so in this case would be
MIN_SNAPSHOT_SIZE=5242880  # 5*1024*1024

PR #55, opened 2019-11-26
Confluence/DCA-160-add-estimated-aws-costs-to-the-user-guide

mmizin:


docs/dc-apps-performance-toolkit-user-guide-jira.md
```diff
@@ -38,6 +38,24 @@ You are responsible for the cost of the AWS services used while running this Qui
...
+- EC instance
+- DB instance
+- EFS
+Based on our research results we have next indicators:
+
+_*The prices below are approximate and may vary depending on factors such as (region, scalability, etc.)_
+
+| Instances      | Costs ($/hr)  |
+| ---------      | --------------|          
+| One instance   | ≈ 1 - 1,3     |
```
astashys: I think we should add one more column where we should add an approximate minimum of hours used by this configuration. For example, one node we need at least for 3 hours, 2 nodes for 2 hours and so on. And then we could also add the total minimum.

ometelytsia: I have doubts that we can correctly estimate the minimum hours needed. Or it would be frustrating for a vendor that he uses Stack much more time than mentioned in the documentation.

astashys: But in another case, the cost per hour will not tell the vendor the money he can spend because he has no idea about the approximate time. We can provide the time we spend by our team and emphasize that he will spend more time and cost will be higher and the vendor could predict cost by multiplying it by 2 or 3 for having approximately digit. Also, in the future, we can add this not required question to the checklist and get a statistic of how much vendors spend money on the testing process using AWS and provide this information instead of ours

docs/dc-apps-performance-toolkit-user-guide-confluence.md
```diff
@@ -38,6 +38,24 @@ You are responsible for the cost of the AWS services used while running this Qui
 
 To reduce costs, we recommend you to keep your deployment up and running only during the performance runs.
 
+### AWS cost estimation ###
```
ometelytsia: [SIMPLE MONTHLY CALCULATOR](https://calculator.s3.amazonaws.com/index.html) provides an estimate of usage charges for AWS services based on certain information you provide.
Monthly charges will be based on your actual usage of AWS services, and may vary from the estimates the Calculator has provided.

_*The prices below are approximate and may vary depending on factors such as (region, instance type, deployment type of DB, etc.)_
| Confluence DC           | Costs ($/hr) |
| ------------------------| -------------|           
| One Node Confluence DC  | ≈ 1,5 - 1,6  |
| Two Node Confluence DC  | ≈ 1,7 - 2,5  |
| Four Node Confluence DC | ≈ 3,5 - 4,2  |

mmizin: Nice!
It seems good to me

docs/dc-apps-performance-toolkit-user-guide-jira.md
ometelytsia: Approximate timings are specified in User Guides. I don't like the idea of putting any numbers here because those numbers may vary a lot depending on too many paraments which we can not control.

astashys: Okay

mmizin: Yep! I'm afraid that I may not have included all the parameters that we use in our stack, so currently, this estimation very approximate!

PR #58, opened 2019-12-02
DCA-250 Create statistic collector script

SergeyMoroz0703:
Python script `statistic_collector.py` will be run after each `bzt` execution in case if `allow_analytics` enabled (by default) in .yml configuration files.
It  will not collect any PI data.

### DIFF
--- a/app/util/analytics.py
+++ b/app/util/analytics.py
@@ -157,7 +157,7 @@ class AnalyticsSender:
                       f'concurrency={self.run_analytics.concurrency}'
 
         r = requests.get(url=f'{base_url}{params_string}')
-        return r.content
+
 
 if __name__ == '__main__':
     app_type = application_type()
@@ -165,4 +165,4 @@ if __name__ == '__main__':
     if p.is_analytics_enabled():
         p.generate_statistics()
         sender = AnalyticsSender(p)
-        print(sender.send_request())
+        sender.send_request()
--- a/app/util/analytics.py
+++ b/app/util/analytics.py
@@ -61,7 +61,7 @@ class AnalyticsFormer:
     def __last_log_dir(self):
         results_dir = None
         results_dirs_dict = {'jira': JIRA_RESULT_DIR, 'confluence': CONFLUENCE_RESULT_DIR}
-        for app, res_dir in results_dirs_dict:
+        for app, res_dir in results_dirs_dict.items():
             if app == self.application_type.lower():
                 results_dir = res_dir
         try:
--- a/app/util/analytics.py
+++ b/app/util/analytics.py
@@ -35,7 +35,7 @@ def application_type():
     return sys.argv[1]
 
 
-class AnalyticsFormer:
+class AnalyticsCollector:
 
     def __init__(self, application_type):
         self.application_type = application_type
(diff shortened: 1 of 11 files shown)


app/util/project_paths.py
```diff
@@ -5,6 +5,13 @@ def __get_jira_yml():
     return Path(__file__).parents[1] / "jira.yml"
 
 
+def __get_jira_result_dir():
+    return Path(__file__).parents[1] / "results" / "jira"
```
astashys: you can make 1 method name it for example `get_artifacts_dir` and `use return os.environ.get('TAURUS_ARTIFACTS_DIR')` Also reuse it in conftest.py script

app/util/analytics.py
```diff
@@ -0,0 +1,172 @@
...
+# List in value in case of specific output appears for some OS for command platform.system()
+OS = {'macOS': ['Darwin'], 'Windows': ['Windows'], 'Linux': ['Linux']}
+DT_REGEX = r'(\d{4}-\d{1,2}-\d{1,2}\s+\d{1,2}:\d{1,2}:\d{1,2})'
+
+BASE_URL = 'http://dcapps-ua-test.s3.us-east-2.amazonaws.com/stats.html?'
+DEV_BASE_URL = 'http://dcapps-ua-test.s3.us-east-2.amazonaws.com/stats_dev.html?'
+
+
+def __validate_app_type():
+    msg = 'Please run util/analytics.py with application type as argument. E.g. python util/analytics.py jira'
```
astashys: can be constant on top level

app/util/analytics.py
```diff
@@ -0,0 +1,172 @@
...
+                      f'exp_dur={self.run_analytics.duration}&act_dur={self.run_analytics.actual_duration}&' \
+                      f'sel_count={self.run_analytics.selenium_test_count}&jm_count={self.run_analytics.jmeter_test_count}&' \
+                      f'concurrency={self.run_analytics.concurrency}'
+
+        r = requests.get(url=f'{base_url}{params_string}')
+
+
+if __name__ == '__main__':
+    app_type = application_type()
+    p = AnalyticsFormer(app_type)
```
astashys: maybe we can change the name `p` to something more understandable?

app/util/analytics.py
```diff
@@ -0,0 +1,172 @@
...
+    except IndexError:
+        SystemExit(msg)
+
+
+def application_type():
+    __validate_app_type()
+    return sys.argv[1]
+
+
+class AnalyticsFormer:
```
astashys: What do you think about `AnalyticsCollector`?

app/util/analytics.py
```diff
@@ -0,0 +1,171 @@
...
+def __validate_app_type():
+    try:
+        app_type = sys.argv[1]
+        if app_type.lower() not in [JIRA, CONFLUENCE, BITBUCKET]:
+            raise SystemExit(APP_TYPE_MSG)
+    except IndexError:
+        SystemExit(APP_TYPE_MSG)
+
+
+def application_type():
```
ometelytsia: let's rename to `get_application_type`

astashys:
Did you try negative cases, for example, if you provided an inexistent host or when Jira server in captcha mode?

PR #66, opened 2019-12-09
Switch analytics collector to lambda api gateway logic

SergeyMoroz0703:


app/util/analytics.py
```diff
@@ -147,20 +143,23 @@ def __init__(self, analytics_instance):
...
-
-        r = requests.get(url=f'{base_url}{params_string}')
-        if r.status_code != 403:
+        headers = {"Content-Type": "application/json"}
+        params = {"date": self.run_analytics.date,
+                  "time_stamp": datetime.now().timestamp(),
+                  "app_type": self.run_analytics.application_type,
+                  "os": self.run_analytics.os,
+                  "tool_ver": self.run_analytics.tool_version,
+                  "run_id": uuid.uuid1().hex,
```
ometelytsia: Let's move run_id calculation to the constructor?

app/util/analytics.py
```diff
@@ -130,11 +123,14 @@ def get_actual_test_count(self):
             elif selenium_test in line:
                 self.selenium_test_count = self.selenium_test_count + 1
 
+    def __convert_to_sec(self, time):
+        seconds_per_unit = {"s": 1, "m": 60, "h": 3600, "d": 86400, "w": 604800}
+        return int(str(time)[:-1]) * seconds_per_unit[str(time)[-1]]
```
ometelytsia: Just int value is also possible and valid format, e.g. `duration: 300` will be the same as `duration: 300s`

app/util/analytics.py
```diff
@@ -39,7 +40,7 @@ class AnalyticsCollector:
 
     def __init__(self, application_type):
         self.application_type = application_type
-        self.run_id = ""
+        self.run_id = uuid.uuid1().hex
```
astashys: I think `uuid.uuid1()` will be enough. UUID is a very popular solution for ID in DB and it is used in default presentation.

app/util/analytics.py
```diff
@@ -130,37 +124,47 @@ def get_actual_test_count(self):
...
-                        f'exp_dur={self.run_analytics.duration}&' \
-                        f'act_dur={self.run_analytics.actual_duration}&' \
-                        f'sel_count={self.run_analytics.selenium_test_count}&' \
-                        f'jm_count={self.run_analytics.jmeter_test_count}&' \
-                        f'concurrency={self.run_analytics.concurrency}'
-
-        r = requests.get(url=f'{base_url}{params_string}')
-        if r.status_code != 403:
+        headers = {"Content-Type": "application/json"}
+        params = {"date": self.analytics.date,
```
astashys: Name `params` can be confusing in this context. I suggest using `body` for example.

astashys: For readability let's order these params. I suggest starting from run_id and time_stamp.

app/util/analytics.py
```diff
@@ -130,37 +124,47 @@ def get_actual_test_count(self):
...
+                  "tool_ver": self.analytics.tool_version,
+                  "run_id": self.analytics.run_id,
+                  "exp_dur": self.analytics.duration,
+                  "act_dur":  self.analytics.actual_duration,
+                  "sel_count": self.analytics.selenium_test_count,
+                  "jm_count": self.analytics.jmeter_test_count,
+                  "concurrency": self.analytics.concurrency
+        }
+        r = requests.post(url=f'{BASE_URL}', json=params, headers=headers)
+        response = r.json()
```
astashys: can confusing, let's name it `response_body`

app/util/analytics.py
```diff
@@ -130,37 +124,46 @@ def get_actual_test_count(self):
...
-                        f'act_dur={self.run_analytics.actual_duration}&' \
-                        f'sel_count={self.run_analytics.selenium_test_count}&' \
-                        f'jm_count={self.run_analytics.jmeter_test_count}&' \
-                        f'concurrency={self.run_analytics.concurrency}'
-
-        r = requests.get(url=f'{base_url}{params_string}')
-        if r.status_code != 403:
+        headers = {"Content-Type": "application/json"}
+        params = {"date": self.analytics.date,
+                  "time_stamp": datetime.now().timestamp(),
```
astashys: This timestamp is float where digits before the coma are seconds. I think we do not need float numbers in our DB. Milliseconds will be a good choice so I suggest `int(round(datetime.utcnow().timestamp() * 1000))`

Also, I suggest don't do redundant operations and reuse variables:
``` 
       utc_now = datetime.utcnow()
        self.time_stamp = int(round(utc_now.timestamp() * 1000))
        self.date = utc_now.strftime("%m/%d/%Y-%H:%M:%S")
```

app/util/analytics.py
SergeyMoroz0703: `response` and `r` are the typical name usage for this case in Python (according to multiple documentation and examples). Do not think that it might be confusable.

app/util/analytics.py
SergeyMoroz0703: I've checked docs, `payload` will be better, will use it

app/util/analytics.py
astashys: I randomly looked at the python code on the internet:
https://www.geeksforgeeks.org/get-post-requests-using-python/
https://realpython.com/python-requests/
https://python-scripts.com/requests

In some cases JSON as `data` in other `json_response` but not found `response` usage for JSON data. But `response` name is used as the name for Http response in some cases but in some cases as `r`. I am ok with `r`.
But using variable name: `response` for JSON in the scope where you have real Http response is confusing. A real response consists not only from JSON data, but it also has other data, also some people using `response` instead of `r` so I think it is very confusing to use `response` for JSON data.

PR #67, opened 2019-12-10
None/switch analytics collector to lambda gateway logic

SergeyMoroz0703:


app/util/analytics.py
```diff
@@ -121,46 +118,88 @@ def get_actual_run_time(self):
...
+                if trigger in line and ('FAIL' in line or 'OK' in line):
+                    test_name = re.findall(test_search_regx, line)[0]
+                    test_rate = float(''.join(re.findall(SUCCESS_TEST_RATE_REGX, line))[:-1])
+                    if test_name not in tests:
+                        tests[test_name] = test_rate
+            return tests
+
+    def get_success_count_from_tests(self, tests):
+        success_test_count = 0
+        for _, success_rate in tests.items():
```
ometelytsia: `for success_rate` in tests.values():

app/util/analytics.py
```diff
@@ -121,46 +118,88 @@ def get_actual_run_time(self):
...
+                success_test_count = success_test_count + 1
+        return success_test_count
+
+
+    def set_actual_test_count(self):
+        TEST_RESULTS_START_STRING = 'Request label stats:'
+        res_string_idx = [index for index, value in enumerate(self.bzt_log_file) if TEST_RESULTS_START_STRING in value]
+        # Cut bzt.log from the 'Request label stats:' string to the end
+        if res_string_idx:
+            res_string_idx = int(res_string_idx[0])
```
ometelytsia: The list should already contain `int` value of `index`. It would be better to calc it on line 147.

app/util/analytics.py
```diff
@@ -5,18 +5,22 @@
 from datetime import datetime
 import hashlib
```
astashys: seems like unused import

PR #71, opened 2019-12-13
fix DCA-277 custom port, remove referer

ometelytsia:


app/jmeter/confluence.jmx
```diff
@@ -61,7 +61,7 @@
         <collectionProp name="HeaderManager.headers">
           <elementProp name="" elementType="Header">
             <stringProp name="Header.name">Origin</stringProp>
-            <stringProp name="Header.value">${application.protocol}://${application.hostname}</stringProp>
+            <stringProp name="Header.value">${application.protocol}://${application.hostname}:${application.port}</stringProp>
```
ometelytsia: The real problem was a missing application port in Origin Header Manager.

PR #99, opened 2020-01-20
limit number of projects for jira

ometelytsia:
Issue found by the vendor.
If project count is more then ~1200 projects (depending on project key length), next request get issues fails with 400 error "Request header is too large".

### DIFF
--- a/app/util/data_preparation/jira/prepare-data.py
+++ b/app/util/data_preparation/jira/prepare-data.py
@@ -79,7 +79,7 @@ def __write_to_file(file_path, items):
 def __create_data_set(jira_api):
     dataset = dict()
     dataset[USERS] = __get_users(jira_api)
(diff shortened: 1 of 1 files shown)


app/util/data_preparation/jira/prepare-data.py
```diff
@@ -120,8 +121,8 @@ def __get_software_project_keys(jira_api):
     software_project_keys = [project['key'] for project in all_projects if 'software' == project.get('projectTypeKey')]
     if not software_project_keys:
         raise SystemExit("There is no software project in Jira")
-
-    return software_project_keys
+    # Limit number of projects to avoid "Request header is too large" for further requests.
+    return software_project_keys[:PROJECTS_COUNT_LIMIT]
```
astashys: As far as I remember, there is the same request via the POST method. Maybe we should try to use it?

PR #105, opened 2020-01-21
Remove w3c timings

ometelytsia:
@jkim2-atlassian couple of vendors are complaining about "InvalidSessionIdException" allover Selenium actions. It is very difficult to reproduce this issue locally, but I was able to see a couple of those failures on my local Mac and on local Docker. I was not able to understand why this had happened, but it looks like some chromedriver/Chrome versions combination flaky issue. And the problem is in this get_w3c_timing method. As we are not using these stats anyway, I propose to remove this code to reduce customers' flakiness rate. Any thoughts?

### DIFF
--- a/app/selenium_ui/conftest.py
+++ b/app/selenium_ui/conftest.py
@@ -39,13 +39,10 @@ def __get_current_results_dir():
 current_results_dir = __get_current_results_dir()
 selenium_results_file = Path(current_results_dir + '/selenium.jtl')
 selenium_error_file = Path(current_results_dir + '/selenium.err')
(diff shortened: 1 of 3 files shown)


jkim2-atlassian:
@ometelytsia we currently don't have any toolings but the idea was to create waterfall charts at end of the test, showing the breakdown. If this is causing flakiness, do remove it. We will reassess at a later time.

ometelytsia:
Ticket to revisit w3c timings later https://bulldog.internal.atlassian.com/browse/DCA-316

PR #110, opened 2020-01-30
Jira/docs update

ometelytsia:
- added KeepAlive flag to avoid broken pipe error when running dataset prepare scritps

- new default db type

- extended reindex instructions as per Approval team request

### DIFF
--- a/docs/dc-apps-performance-toolkit-user-guide-jira.md
+++ b/docs/dc-apps-performance-toolkit-user-guide-jira.md
@@ -73,6 +73,7 @@ The Data Center App Performance Toolkit officially supports:
 | Cluster node instance type | [c5.4xlarge](https://aws.amazon.com/ec2/instance-types/c5/) |
 | Maximum number of cluster nodes | 1 |
 | Minimum number of cluster nodes | 1 |
+| Cluster node instance volume size | 100 |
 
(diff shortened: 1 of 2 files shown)


docs/dc-apps-performance-toolkit-user-guide-jira.md
```diff
@@ -82,7 +83,7 @@ The Data Center App Performance Toolkit framework is also set up for concurrency
 
 | Parameter | Recommended Value |
 | --------- | ----------------- |
-| Database instance class | [db.m4.large](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html#Concepts.DBInstanceClass.Summary) |
+| Database instance class | [db.m5.large](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html#Concepts.DBInstanceClass.Summary) |
```
benmagro: What are we making an instance size change? Does this effect the costs?

ometelytsia: AWS moved to a new generation of DB instances. So there is no way to select db.m4.large from the dropdown during quickstart wizard (only from CLI command).

Price: 
db.m4.large - $0.35 per hour for multi az deployment
db.m5.large - $0.342 per hour for multi az deployment

Confluence template was not updated yet, so it still uses db.m4.large.

PR #113, opened 2020-02-03
DCA-323 user git configuration

mmizin:


app/jmeter/bitbucket.jmx
```diff
@@ -478,6 +478,47 @@ String gitTmpWorkspace = props.get(&quot;GIT_TMP_WORKSPACE&quot;);
...
+        BufferedReader reader = new BufferedReader(
+                new InputStreamReader(p.getInputStream()));
+
+        String line;
+        boolean name = false;
+        boolean email = false;
+
+        while ((line = reader.readLine()) != null) {
+            if (line.contains(&quot;user.name&quot;)) {
+                if (line.endsWith(&quot;=&quot;)) {
```
astashys: You just check if a user name is empty but It seems like it will not work if it is " " or "      ".
Also read all config is redundant.
You can execute `git config --global user.name` and `git config --global user.email` to get the value and check only them separately instead of iterating over the whole config. In this case please use `org.apache.commons.lang3.StringUtils.isBlank()` Jmeter has this lib in classpath. But for local testing you can download it from here https://mvnrepository.com/artifact/org.apache.commons/commons-lang3

mmizin: Thx for the advice!!!

PR #115, opened 2020-02-04
DCA-321 Git as requirement and README for bitbucket

ometelytsia:
DCA-321


astashys:
We do not have bitbucket changes in dev yet, what do you think about changing destination branch to bitbucket branch?

PR #120, opened 2020-02-07
[BB JMeter] Do not remove all ssh keys from admin user when adding new

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-314


astashys:
please remove not needed changes from your PR, please

PR #130, opened 2020-02-12
DCA-342: Handle permission issue on Bitbucket

astashys:
https://bulldog.internal.atlassian.com/browse/DCA-342

### DIFF
--- a/app/util/data_preparation/api/bitbucket_clients.py
+++ b/app/util/data_preparation/api/bitbucket_clients.py
@@ -1,8 +1,18 @@
+from enum import Enum
+
 from util.data_preparation.api.abstract_clients import RestClient
 
 BATCH_SIZE_PROJECTS = 100
 BATCH_SIZE_USERS = 100
 BATCH_SIZE_REPOS = 100
+
+
+class BitbucketUserPermission(Enum):
+    LICENSED_USER = "LICENSED_USER"
+    PROJECT_CREATE = "PROJECT_CREATE"
+    ADMIN = "ADMIN"
+    SYS_ADMIN = "SYS_ADMIN"
+
 
 class BitbucketRestClient(RestClient):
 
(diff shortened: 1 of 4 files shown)


app/util/data_preparation/api/bitbucket_clients.py
```diff
@@ -23,6 +33,7 @@ def get_entities(self, entity_name, batch_size, filter=None, max_results=500):
             start_at = start_at + len(response.json()['values'])
             loop_count -= 1
             if loop_count == 1:
+                # TODO variable max_results is not used, before delete it we should check if there is no bug in the code
```
ometelytsia: pls create a Jira ticket and add an id to this TODO

app/util/data_preparation/bitbucket/prepare-data.py
```diff
@@ -19,15 +19,31 @@ def generate_random_string(length=20):
...
     while perf_user_count_to_create > 0:
-        user = bitbucket_api.create_user(username=f'{DEFAULT_USER_PREFIX}-{generate_random_string(5)}')
-        perf_users.append(user)
+        client.create_user(f'{DEFAULT_USER_PREFIX}-{generate_random_string(5)}')
         perf_user_count_to_create = perf_user_count_to_create - 1
-    return bitbucket_api.get_users(username=f'{DEFAULT_USER_PREFIX}', max_results=perf_user_count)
+
+    perf_users = client.get_users(f'{DEFAULT_USER_PREFIX}', perf_users_desired_number)
+    if len(perf_users) < perf_users_desired_number:
+        raise SystemExit(f'Server returned less number of users than expected')
```
ometelytsia: we could add variables into log message

app/util/data_preparation/api/bitbucket_clients.py
astashys: I am going to investigate it today

app/util/data_preparation/bitbucket/prepare-data.py
astashys: it is a very rare case and we could have the numbers from users in csv file and concurrency in yml file

PR #131, opened 2020-02-13
Dca 203 create bitbucket restore script for deploying to quickstart

SergeyMoroz0703:


app/util/bitbucket/upload_attachments.sh
```diff
@@ -0,0 +1,82 @@
...
+FREE_SPACE_GB=$((${FREE_SPACE_KB}/1024/1024))
+REQUIRED_SPACE_GB=$((5 + ${ARTIFACT_SIZE_GB}))
+if [[ ${FREE_SPACE_GB} -lt ${REQUIRED_SPACE_GB} ]]; then
+   echo "Not enough free space for download."
+   echo "Free space: ${FREE_SPACE_GB} GB"
+   echo "Required space: ${REQUIRED_SPACE_GB} GB"
+   exit 1
+fi;
+
+wget -qO- ${ATTACHMENTS_TAR_URL}| tar -xvz -C ${ATTACHMENT_DIR_DATA} --strip-components 1
```
ometelytsia: Why do we need `--strip-components 1`?

ometelytsia: Could you pls add `time` to this command.
This help us to track and improve timings in future if needed.

ometelytsia: Looks like tar with `-v` output all unarchived file names into the log file.
As we have a huge amount of files, pls check if we could use: --checkpoint=.10000 flag and remove `-v` -> this way tar will print one dot per 10k files.

app/util/bitbucket/upload_attachments.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
ometelytsia: Can we add a check that we are on NFS server, but not on the node.
There is a big chance that the user could start this script on the backend node.

app/util/bitbucket/populate_db.sh
```diff
@@ -0,0 +1,162 @@
...
+    echo "Check correctness of install command or install Postgres client manually."
+    echo "INSTALL_PSQL_CMD=${INSTALL_PSQL_CMD}"
+    exit 1
+  fi
+else
+  echo "Postgres client is already installed"
+fi
+
+echo "Step2: Download DB dump"
+DUMP_DIR='/media/atl/bitbucket/shared'
```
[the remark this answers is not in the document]
SergeyMoroz0703: There is the place where free disk space is available 
`1.2T  500G  700G  42% /media/atl`

app/util/bitbucket/upload_attachments.sh
SergeyMoroz0703: --strip-components 1 allow us to untar without creation additional directory inside the `data` dir.

SergeyMoroz0703: Good point, already tested it, will add.

ometelytsia: why do we have this additional directory?

app/util/bitbucket/populate_db.sh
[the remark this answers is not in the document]
ometelytsia: looks like this folder is on a NFS server

PR #132, opened 2020-02-13
DCA 268 check if api enabled confluence

SergeyMoroz0703:


app/util/data_preparation/api/confluence_clients.py
```diff
@@ -122,6 +122,13 @@ def search(self, cql, cqlcontext=None, expand=None, start=0, limit=500):
 
         return search_results_list
 
+    def is_remote_api_enabled(self):
+        api_url = f'{self.host}/rpc/xmlrpc'
+        response = self.get(api_url, error_msg='Confluence Remote API (XML-RPC & SOAP) is disabled. '
+                                               'For further processing enable Remote API via '
+                                               'General Configuration - Further Configuration - Remote API')
+        return response.status_code == 200
```
astashys: if in a negative case the returned status code is not less than 400, you can just return `True` and don't do redundant checks, because `self.get()` already checks if the status code is less than 400.

SergeyMoroz0703: In this case I need to know that status code is EXACTLY 200, not less than 400. If status code equal to 200, it means that remote api is enabled, otherwise not. That is why I use exactly 200 status code.

PR #138, opened 2020-02-14
DCA-344: Inconsistent work of prepare-data script with users

astashys:
https://bulldog.internal.atlassian.com/browse/DCA-344

### DIFF
--- a/app/util/data_preparation/api/bitbucket_clients.py
+++ b/app/util/data_preparation/api/bitbucket_clients.py
@@ -1,3 +1,4 @@
+import time
 from enum import Enum
 
 from util.data_preparation.api.abstract_clients import RestClient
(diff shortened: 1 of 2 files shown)


app/util/data_preparation/api/bitbucket_clients.py
```diff
@@ -57,7 +64,7 @@ def get_projects(self, max_results=500):
 
     def get_users(self, name_filter, max_results=500):
         return self.get_entities(entity_name='users',
-                                 filter=name_filter,
+                                 filter_=name_filter,
```
astashys: I updated the name of this parameter since it hides built-in function `filter`: https://docs.python.org/2/library/functions.html
According to this doc: https://dbader.org/blog/meaning-of-underscores-in-python there is a convention to add underscore after the name in this case.

PR #159, opened 2020-03-06
DCA-332: Add a check if git was installed

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-322?workflowName=Software+Simplified+Workflow+for+Project+DA&stepId=1


app/util/git_client_check.py
```diff
@@ -0,0 +1,7 @@
+from subprocess import check_output
+
+
+try:
+    print("Git version: {}".format(check_output(["git", "--version"])))
+except Exception:
+    raise Exception("Please check if git installed and available from command line.")
```
astashys: Do we really support all git versions? Is `git clone --no-checkout` supported from the first git version?

SergeyMoroz0703: `git` version 1.8.2 has `--no-checkout` flag, checked

ometelytsia: Good question.
@SergeyMoroz0703+++++

PR #164, opened 2020-03-24
None refactor analytics datetime iso8601 format

SergeyMoroz0703:


ometelytsia:
Pls check why CI is red before merge.

PR #165, opened 2020-03-25
None pageobject pattern bitbucket

SergeyMoroz0703:


app/selenium_ui/bitbucket/pages/pages.py
```diff
@@ -0,0 +1,360 @@
...
+    def __init__(self, driver, project_key, repo_slug):
+        BasePage.__init__(self, driver)
+        url_manager = UrlManager(project_key=project_key, repo_slug=repo_slug)
+        self.page_url = url_manager.repo_url()
+        self.params_to_verify = url_manager.repo_params_browse
+        self.repo_slug = repo_slug
+        self.project_key = project_key
+
+    def at(self):
+        return self.verify_url(self.params_to_verify)
```
ometelytsia: Did not get why every class has the same method?

PR #168, opened 2020-04-01
NONE pageobject pattern jira

SergeyMoroz0703:


app/selenium_ui/jira/pages/pages.py
```diff
@@ -0,0 +1,196 @@
...
+
+from selenium_ui.base_page import BasePage
+from selenium_ui.jira.pages.selectors import UrlManager, LoginPageLocators, DashboardLocators, PopupLocators, \
+    IssueLocators, ProjectLocators, SearchLocators, BoardsListLocators, BoardLocators, LogoutLocators
+
+
+class PopupManager(BasePage):
+
+    def dismiss_default_popup(self):
+        return self.dismiss_popup(PopupLocators.default_popup, PopupLocators.popup_1, PopupLocators.popup_2)
```
ometelytsia, line 13: maybe it worth to have user-friendly names for popups instead of popup_1?

SergeyMoroz0703: Good catch, I'll try to find popups with these selectors and will rename it in further pull request

PR #169, opened 2020-04-01
Dca-318 results validation

ometelytsia:
Generate a new file under the results folder: `results_summary.log`.
File contains a detailed summary of the run with a user-friendly error messages if any.

Example:
image


app/util/analytics.py
```diff
@@ -208,6 +223,135 @@ def generate_analytics(self):
         self.set_date_timestamp()
         self.application_version = self.get_application_version()
 
+    @property
+    def actual_git_operations_count(self):
+        count = 0
+
+        if self.application_type != BITBUCKET:
```
SergeyMoroz0703, line 230: Not sure why do we need this check considering the fact this function is called after additional checks on the 317, 329 rows

ometelytsia: extra protection in case later on someone will decide to use this property

PR #178, opened 2020-04-07
 DCA-415: Change get request method on post

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-415

### DIFF
--- a/app/util/data_preparation/api/jira_clients.py
+++ b/app/util/data_preparation/api/jira_clients.py
@@ -1,3 +1,5 @@
+from typing import List
+
 from util.data_preparation.api.abstract_clients import RestClient
 
 BATCH_SIZE_BOARDS = 1000
@@ -76,13 +78,12 @@ class JiraRestClient(RestClient):
 
         return users_list
 
-    def issues_search(self, jql='order by key', start_at=0, max_results=1000, validate_query=True, fields='id'):
+    def issues_search(self, jql='order by key', start_at=0, max_results=1000, fields=['id']):
         """
         Searches for issues using JQL.
         :param jql: a JQL query string
         :param start_at: the index of the first issue to return (0-based)
         :param max_results: the maximum number of issues to return (defaults to 50).
-        :param validate_query: whether to validate the JQL query
         :param fields: the list of fields to return for each issue. By default, all navigable fields are returned.
         *all - include all fields
         *navigable - include just navigable fields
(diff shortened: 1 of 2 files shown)


app/util/data_preparation/api/jira_clients.py
```diff
@@ -97,9 +98,15 @@ def issues_search(self, jql='order by key', start_at=0, max_results=1000, valida
         last_loop_remainder = max_results % BATCH_SIZE_ISSUES
 
         while loop_count > 0:
-            api_url = f'{self.host}/rest/api/2/search?jql={jql}&startAt={start_at}&maxResults={max_results}' \
-                      f'&validateQuery={validate_query}&fields={fields}'
-            response = self.get(api_url, "Could not retrieve issues")
+            api_url = f'{self.host}/rest/api/2/search'
```
ometelytsia, line 101: api_url variable is not changing in loop cycles, so it does not make sense to re-assign it on every loop. Let's move out it from a `while` statement.

app/util/data_preparation/api/jira_clients.py
```diff
@@ -97,9 +98,15 @@ def issues_search(self, jql='order by key', start_at=0, max_results=1000, valida
...
+            pay_load = {
+                    "jql": f"{jql}",
```
ometelytsia, line 103: No need to use f-annotation here. Just `"jql": jql,`

app/util/data_preparation/api/jira_clients.py
```diff
@@ -97,9 +98,15 @@ def issues_search(self, jql='order by key', start_at=0, max_results=1000, valida
         last_loop_remainder = max_results % BATCH_SIZE_ISSUES
 
         while loop_count > 0:
-            api_url = f'{self.host}/rest/api/2/search?jql={jql}&startAt={start_at}&maxResults={max_results}' \
-                      f'&validateQuery={validate_query}&fields={fields}'
-            response = self.get(api_url, "Could not retrieve issues")
+            api_url = f'{self.host}/rest/api/2/search'
+            pay_load = {
```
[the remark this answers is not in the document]
mmizin: I will rename itto body

app/util/data_preparation/api/jira_clients.py
```diff
@@ -92,14 +91,23 @@ def issues_search(self, jql='order by key', start_at=0, max_results=1000, valida
         :return: Returns the requested issues
         """
 
+        if fields is None:
```
astashys, line 94: you can set it as default value in the function signature

app/util/data_preparation/api/jira_clients.py
```diff
@@ -92,14 +91,23 @@ def issues_search(self, jql='order by key', start_at=0, max_results=1000, valida
...
         while loop_count > 0:
-            api_url = f'{self.host}/rest/api/2/search?jql={jql}&startAt={start_at}&maxResults={max_results}' \
-                      f'&validateQuery={validate_query}&fields={fields}'
-            response = self.get(api_url, "Could not retrieve issues")
+
+            body = {
+                    "jql": jql,
+                    "startAt": start_at,
+                    "maxResults": max_results,
+                    "fields": fields
```
astashys, line 107: "fields": ['id'] if fields is None else fields

PR #181, opened 2020-04-13
DCA-187: Improve validation logic in csv aggregator scrtipt

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-187


ometelytsia:
Pls do not forget to fix linter comments.

PR #183, opened 2020-04-21
Dca 297 fix flake8 python pep warnings

SergeyMoroz0703:
Comment  ` # noqa` means skip specific pep8 code for a particular string
http://flake8.pycqa.org/en/3.2.0/user/ignoring-errors.html


app/selenium_ui/base_page.py
```diff
@@ -110,7 +110,7 @@ def dismiss_popup(self, *args):
         for elem in args:
             try:
                 self.driver.execute_script(f"document.querySelector(\'{elem}\').click()")
-            except:
+            except:  # noqa E722
```
ometelytsia, line 113: What do you think about this workaround:
```
from selenium.common.exceptions import WebDriverException
try:
    self.driver.execute_script(f"document.querySelector(\'{elem}\').click()")
except(WebDriverException, Exception):
    pass
```

app/selenium_ui/conftest.py
```diff
@@ -254,5 +253,5 @@ def __call__(self, w_driver):
             try:
                 if fn(w_driver):
                     return True
-            except:
+            except:  # noqa E722
```
ometelytsia, line 256: The same as above.

PR #185, opened 2020-04-22
DCA-350 skip whole tests if login failed

SergeyMoroz0703:
Screen Shot 2020-04-22 at 10 14 37 PM
Screen Shot 2020-04-22 at 10 14 58 PM
Screen Shot 2020-04-22 at 11 01 09 PM

Each time login is failed, all next actions will be skipped. 
In the picture above - an example of the current implementation for Confluence. Failed login does not affect other actions.

### DIFF
--- a/app/selenium_ui/conftest.py
+++ b/app/selenium_ui/conftest.py
@@ -24,7 +24,7 @@ SCREEN_HEIGHT = 1080
 
 JTL_HEADER = "timeStamp,elapsed,label,responseCode,responseMessage,threadName,success,bytes,grpThreads,allThreads," \
              "Latency,Hostname,Connect\n"
-
+LOGIN_ACTION_NAME = 'login'
 
 def __get_current_results_dir():
     if 'TAURUS_ARTIFACTS_DIR' in os.environ:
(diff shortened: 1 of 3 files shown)


app/selenium_ui/conftest.py
```diff
@@ -58,6 +58,12 @@ def print_timing(interaction=None):
     def deco_wrapper(func):
         @functools.wraps(func)
         def wrapper():
+            global fail_login
+            if LOGIN_ACTION_NAME in interaction:
+                fail_login = False
```
ometelytsia, line 63: fail_login sound like a method to fail login. Let's change the variable name to 'login_failed'.

app/selenium_ui/conftest.py
```diff
@@ -58,6 +58,12 @@ def print_timing(interaction=None):
...
+            if fail_login:
+                pytest.skip(f"login is failed")
+            fail_login = False
```
ometelytsia, line 66: It looks like this like is not needed. Need to remove.

app/selenium_ui/conftest.py
```diff
@@ -58,6 +58,12 @@ def print_timing(interaction=None):
...
             f_interaction = interaction
```
ometelytsia, line 67: let's remove this line, and use `interaction` everywhere.

PR #189, opened 2020-04-27
DCA-433-added dataset information to results summary log

SergeyMoroz0703:


app/util/analytics.py
```diff
@@ -357,6 +387,7 @@ def generate_report_summary(self):
         summary_report.append(f'OS|{self.os}')
         summary_report.append(f'DC Apps Performance Toolkit version|{self.tool_version}')
         summary_report.append(f'Application|{self.application_type} {self.application_version}')
+        summary_report.append(f'Dataset info|{self.get_dataset_information()}')
```
ometelytsia, line 390: looks like you need to you here self.dataset_information variable, or remove line 286.

PR #196, opened 2020-04-30
DCA-452: Check product language before tests start

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-452

### DIFF
--- a/app/util/data_preparation/api/bitbucket_clients.py
+++ b/app/util/data_preparation/api/bitbucket_clients.py
@@ -1,5 +1,6 @@
 import time
 from enum import Enum
+from bs4 import BeautifulSoup
 
 from util.data_preparation.api.abstract_clients import RestClient
 
@@ -140,3 +141,9 @@ class BitbucketRestClient(RestClient):
         headers['Origin'] = self.host
         r = session.post(url, data=body, headers=headers)
         return r.content.decode('utf-8')
+
+    def get_locale(self):
+        response = self.get(self.host, f'Could not retrieve page')
+        soup = BeautifulSoup(response.text, 'html.parser')
+        current_lang = soup.html['lang']
+        return current_lang
--- a/app/util/data_preparation/confluence/prepare-data.py
+++ b/app/util/data_preparation/confluence/prepare-data.py
@@ -90,6 +90,13 @@ def write_test_data_to_files(dataset):
     __write_to_file(CONFLUENCE_USERS, users)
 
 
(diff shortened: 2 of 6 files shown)


SergeyMoroz0703:
You use BeautifulSoup to parse `html` responses. This is not a built-in Python library, so if you use third-party library you should add it to `requirements.txt` file. 
On the other hand, I **do not** think we need to implement BeatifulSoup as an additional `html` parser, all pages could be parsed by using built-in Python library such as lxml. Please review this implementation and use built-in methods for `html` parsing to avoid using additional libraries.

app/util/data_preparation/confluence/prepare-data.py
```diff
@@ -90,6 +90,13 @@ def write_test_data_to_files(dataset):
     __write_to_file(CONFLUENCE_USERS, users)
 
 
+def __check_default_confluence_language(confluence_api):
+    confluence_language = confluence_api.get_locale()
```
SergeyMoroz0703, line 94: All this code is situated in the `confluence` directory and all these methods apply just to Confluence. So using `confluence` in the variables naming is redundant. `current_language` is better.

app/util/data_preparation/confluence/prepare-data.py
```diff
@@ -90,6 +90,13 @@ def write_test_data_to_files(dataset):
...
+    if confluence_language != 'en_US':
+        raise SystemExit(f'Unfortunately, we do not support {confluence_language} language! '
```
SergeyMoroz0703, line 96: `f'{confluence_language} language is not supported. Please change your profile language **to** English'`

app/util/data_preparation/confluence/prepare-data.py
mmizin: done

app/util/data_preparation/confluence/prepare-data.py
mmizin: done

ometelytsia:
Just to confirm, did you test on all versions of all products?

app/util/data_preparation/api/bitbucket_clients.py
```diff
@@ -2,6 +2,7 @@
 from enum import Enum
 
 from util.data_preparation.api.abstract_clients import RestClient
+import lxml.html as LH
```
astashys, line 5: seems like PEP8 violation. N813  camelcase imported as lowercase

PR #198, opened 2020-05-07
DCA-457: Add a collaborative editing status check

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-457

### DIFF
--- a/app/util/data_preparation/api/confluence_clients.py
+++ b/app/util/data_preparation/api/confluence_clients.py
@@ -139,6 +139,11 @@ class ConfluenceRestClient(RestClient):
         response = self.get(api_url, 'Could not get issues count')
         return response.json().get('totalSize', 0)
 
+    def get_collaborative_editing_status(self):
+        api_url = f'{self.host}/rest/synchrony-interop/status'
+        response = self.get(api_url, error_msg='Could not get collaborative editing status').json()
+        return response
+
 
 class ConfluenceRpcClient(Client):
 
--- a/app/util/data_preparation/api/confluence_clients.py
+++ b/app/util/data_preparation/api/confluence_clients.py
@@ -144,7 +144,7 @@ class ConfluenceRestClient(RestClient):
         api_url = f'{self.host}/rest/synchrony-interop/status'
         response = self.get(api_url, error_msg='Could not get collaborative editing status')
         return response.json()
-      
+
     def get_locale(self):
         page = LH.parse(self.host)
         try:
--- a/app/util/data_preparation/api/confluence_clients.py
+++ b/app/util/data_preparation/api/confluence_clients.py
@@ -2,6 +2,7 @@ import xmlrpc.client
 
 from util.data_preparation.api.abstract_clients import RestClient, Client
 import xml.etree.ElementTree as ET
+import lxml.html as LH
 
 BATCH_SIZE_SEARCH = 500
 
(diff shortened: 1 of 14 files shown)


app/util/data_preparation/api/confluence_clients.py
```diff
@@ -139,6 +139,11 @@ def get_total_pages_count(self):
         response = self.get(api_url, 'Could not get issues count')
         return response.json().get('totalSize', 0)
 
+    def get_collaborative_editing_status(self):
+        api_url = f'{self.host}/rest/synchrony-interop/status'
+        response = self.get(api_url, error_msg='Could not get collaborative editing status').json()
+        return response
+
```
SergeyMoroz0703, line 146: Good catch of `status` endpoint! But I figured out that there is option 'Limited' in Collaborative mode. In this case `Only one person will be able to edit at a time.` (doc). 
What do you think about add force enabling of collaborative editing?
`/rest/synchrony-interop/enable` this endpoint helps in this case, it starts a job for enabling CE(1-2sec).

mmizin: Thx that you mentioned about the "Collaborative editing LIMITED"  option. We have to check the "synchronyEnabled" status too! I will add this check obligatorily.
As regards to force enabling of collaborative editing it is a good idea, that's what I thought too! I think I need to spend some time to investigate would it have a bad impact on us/vendors!

app/util/data_preparation/api/confluence_clients.py
```diff
@@ -139,6 +139,11 @@ def get_total_pages_count(self):
         response = self.get(api_url, 'Could not get issues count')
         return response.json().get('totalSize', 0)
 
+    def get_collaborative_editing_status(self):
+        api_url = f'{self.host}/rest/synchrony-interop/status'
+        response = self.get(api_url, error_msg='Could not get collaborative editing status').json()
```
ometelytsia, line 144: It's better not to mix response and response serialization.
```
response = self.get(api_url, error_msg='Could not get collaborative editing status')
return response.json
```

app/util/data_preparation/confluence/prepare-data.py
```diff
@@ -90,6 +90,13 @@ def write_test_data_to_files(dataset):
     __write_to_file(CONFLUENCE_USERS, users)
 
 
+def __check_collaborative_editing_enable(confluence_api):
```
ometelytsia, line 93: For consistency let's rename method name to `__is_collaborative_editing_enabled`

app/util/data_preparation/confluence/prepare-data.py
```diff
@@ -90,6 +90,13 @@ def write_test_data_to_files(dataset):
...
+    status = confluence_api.get_collaborative_editing_status()
+    if not all(status.values()):
+        raise Exception('Synchrony plugin and/or the collaborative editing feature are disabled in Confluence. '
+                        'Please turn collaborative editing back on.')
```
ometelytsia, line 97: Better wording here would be:
`Please turn on collaborative editing in Confluence System Preferences page in order to run DC Apps Performance Toolkit.`

app/util/data_preparation/api/confluence_clients.py
mmizin: Done

app/util/data_preparation/confluence/prepare-data.py
mmizin: Done

app/util/data_preparation/confluence/prepare-data.py
mmizin: Done

PR #200, opened 2020-05-08
added handling of 404 error

SergeyMoroz0703:


app/util/data_preparation/api/abstract_clients.py
```diff
@@ -86,5 +86,8 @@ def __verify_response(self, response: Response, error_msg: str):
             if denied_reason and denied_reason.startswith('CAPTCHA_CHALLENGE'):
                 raise Exception(f"User name [{self.user}] is in Captcha Mode. " +
                                 "Please login via Web UI first and re-run tests.")
-
+        elif status_code == 404:
+            raise Exception(f"Could not found URL {response.url}. "
```
astashys, line 90: Maybe we can change `Could not found URL {response.url}.` to `Server returned 404(Not Found)`.

SergeyMoroz0703: Good idea. What do you think about `The URL or content are not found for {response.url}`?
I would add URL (404) to let the user debugs easier.

PR #201, opened 2020-05-08
fixed user creation infinity loop

SergeyMoroz0703:


app/util/data_preparation/confluence/prepare-data.py
```diff
@@ -12,9 +12,9 @@
 USERS = "users"
 PAGES = "pages"
 BLOGS = "blogs"
-DEFAULT_USER_PREFIX = 'performance_'
+DEFAULT_USER_PREFIX = 'performanceAA_'
```
mmizin: I think you forgot to change the prefix back

SergeyMoroz0703: already fix it above ^^^ `fix misstypes` commit

PR #209, opened 2020-05-27
Fix bbs run with postfix.

ometelytsia:
* Fix case when number of project/repos is very small
* Fix bbs run with postfix


app/util/data_preparation/api/bitbucket_clients.py
```diff
@@ -52,10 +52,16 @@ def get_non_fork_repos(self, max_results):
...
-            for repo in response.json()['values']:
-                if 'origin' not in repo and len(non_fork_repos) < max_results:
+            repos = response.json()['values']
+            for repo in repos:
+                if 'origin' not in repo:
                     non_fork_repos.append(repo)
-            start_at = start_at + batch_size
+                    if len(non_fork_repos) == max_results:
+                        break
+            if len(repos) == batch_size:
```
astashys, line 61: you can simplify construction by:
if len(repos) != batch_size:
    break
start_at = start_at + batch_size

But anyway it seems like here we can have a bug in the future if somebody changes `batch_size` in our code to the value that exceeds the maximum value that is allowed by API, or if Bitbucket API will change maximum allowed value. For example, when `batch_size` was 1000 but API allows maximum 100, and then API would return 100 items and our code will break the loop because `len(repos) != batch_size`. 
From the documentation: `The limit parameter indicates how many results to return per page. Most APIs default to returning 25 if the limit is left unspecified. This number can be increased, but note that a resource-specific hard limit will apply. These hard limits can be configured by server administrators, so it's always best practice to check the limit attribute on the response to see what limit has been applied`

ometelytsia: refactored, @astashys++++++

PR #224, opened 2020-06-06
Dca 528 locust support for confluence

SergeyMoroz0703:


app/locustio/confluence/requests_params.py
```diff
@@ -0,0 +1,116 @@
+# flake8: noqa
```
ometelytsia, line 1: why do you need this here?

app/locustio/confluence/requests_params.py
```diff
@@ -0,0 +1,116 @@
...
+    data_sets["pages"] = read_input_file(CONFLUENCE_PAGES)
+    data_sets["blogs"] = read_input_file(CONFLUENCE_BLOGS)
+    data_sets["users"] = read_input_file(CONFLUENCE_USERS)
+    data_sets['static-content'] = read_input_file(CONFLUENCE_STATIC_CONTENT)
+
+    return data_sets
+
+
+class BaseResource:
+    resources_file = 'locustio/jira/resources.json'
```
ometelytsia, line 18: 'locustio/jira/resources.json' ?

SergeyMoroz0703: Good catch :) will fix it

app/locustio/confluence/requests_params.py
SergeyMoroz0703: There is a lot of `W605 invalid escape sequence` pep warning in this file in regular expressions.

PR #230, opened 2020-06-10
Added information for using WSL on Windows

scrobbleme:
Using WSL is much easier and less error prone on Windows as the Linux instructions can be used directly.


scrobbleme:
Hope I got it right, I'm not so familiar with Github pull request feature

README.md
```diff
@@ -62,7 +62,18 @@ source venv/bin/activate
...
 ```
 
-### Windows
+### Windows (WSL)
+
+1. Install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10) and a Linux distribution of your choice (Ubuntu is recommended)
+1. Start bash 
+   1. Install dependencies ``sudo apt-get update; sudo apt-get install python3-pip``
+   1. Install Chrome Browser
+      1. ``wget -q -O - https://dl-ssl.google.com/linux/linux_signing_key.pub | sudo apt-key add -``
```
ometelytsia, line 71: Why do you use double ` (backquote) but not single?

scrobbleme: Ah wasn't aware of this. Just thought this is how it works.

ometelytsia: Single backquote should do the trick.

README.md
```diff
@@ -62,7 +62,18 @@ source venv/bin/activate
 pip install -r requirements.txt
 ```
 
-### Windows
+### Windows (WSL)
+
+1. Install [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/install-win10) and a Linux distribution of your choice (Ubuntu is recommended)
+1. Start bash 
+   1. Install dependencies ``sudo apt-get update; sudo apt-get install python3-pip``
```
ometelytsia, line 69: one more place with double backquotes

ometelytsia:
@astashys could you pls test instructions on Windows.

ometelytsia:
Hi @scrobbleme.
Could you pls specify your scenario for WSL usage.

In release 3.0.0 we going to introduce dcapt docker container, and it should cover all possible use cases for WSL - simple run on windows from the console.

For app-specific actions development real browser, IDE and local setup are needed anyway.

Please confirm if docker container will cover main use cases for WSL or you have some more in mind.

Thank you,
Alex

scrobbleme:
@ometelytsia 

My use case:  use Linux within Windows and therefor just run the DC tests much easier. That's it.

I don't know if Docker will cover all things we need, at the end we just need to be able to get the tests running. So if it works out, okay...
With "plain" Windows ---> pain. With WSL --> just followed the instructions and it worked within 30 minutes.

---

We are currently using Docker for Confluence development, but I already switched to WSL as it is much faster and easier to use (for me) and uses less resources than Docker (one layer less since docker uses WSL now too). Also, in WSL I can just access files... from Windows, open logs, edit configs...
So personally, I probably will prefer WSL over Docker in the future...

ometelytsia:
It is totally up to you of course what to use WSL, Docker or local env.

Docker container for DCAPT has everything inside (chrome, python, libs, etc) so only Docker installation is needed. Logs produced would be accessible in Windows and any file could be edited in windows as well.

So I'm declining this PR, for now, to reduce possible confusion for not very technical vendors. 
Also, I'll try to test out WSL approach on vendors which for some reason stuck with local setup and docker to see how it works for them.

PR #232, opened 2020-06-10
more precise selection for comment button

kovetskiy:
Currently, the test searches for `Comment` button by a very abstract selector:
`div.buttons>button:nth-child(1)`.

If an add-on adds a few new buttons before the PR diff and has a `div` with class
`buttons`, it will be selected instead of the comment button.

This pull request changes the selector and makes selenium search for the comment
button exactly in the comment form footer.


ometelytsia:
Hi @kovetskiy. Thank you for your contribution. Could you pls create a new branch from a `dev` branch, then add your improvement and then create a PR from your branch to a deb branch.
In current PR your branch was created from `master`, so there are conflicts and unwanted changes in other files. Thanks.

PR #233, opened 2020-06-10
more precise selection for comment button

kovetskiy:
#232


kovetskiy:
Please add some notes about the branch policy to the readme file, it will make contributions much more convenient.

PR #246, opened 2020-06-24
DCA-629 local jtl to csv convertor

SergeyMoroz0703:


ometelytsia:
@SergeyMoroz0703 Need to remove `jpgc-cmd=2.2` from .yml files.

PR #266, opened 2020-07-07
DCA-603: Check if rte disabled

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-603


app/util/data_preparation/jira_prepare_data.py
```diff
@@ -24,6 +24,9 @@
 
 ENGLISH = 'en_US'
 
+URL = JIRA_SETTINGS.server_url
+CLIENT = JiraRestClient(URL, JIRA_SETTINGS.admin_login, JIRA_SETTINGS.admin_password)
```
SergeyMoroz0703, line 28: Why do you decide to put `CLIENT` outside the `main` function? 
If you do not use `__check_rte_status` function in prepare_data.py, why do you place it here instead if jira_clients?
In this case you can import `client` and `__check_rte_status` from jira_clients?

app/util/data_preparation/jira_prepare_data.py
```diff
@@ -153,17 +156,18 @@ def __check_current_language(jira_api):
...
 
-def main():
-    print("Started preparing data")
-
-    url = JIRA_SETTINGS.server_url
-    print("Server url: ", url)
+def __check_rte_status(jira_api):
+    app_prop = jira_api.get_applications_properties()
+    rte_status = [i['value'] for i in app_prop if i['id'] == 'jira.rte.enabled'][0]
+    return rte_status
```
SergeyMoroz0703, line 162: It would be useful to return boolean instead of string. `return rte_status=='true'`

app/util/api/jira_clients.py
```diff
@@ -195,3 +195,8 @@ def get_locale(self):
         api_url = f'{self.host}/rest/api/2/myself'
         user_properties = self.get(api_url, "Could not retrieve user")
         return user_properties.json()['locale']
+
+    def get_applications_properties(self):
+        api_url = f'{self.host}/rest/api/2/application-properties'
+        user_properties = self.get(api_url, "Could not retrieve user")
```
SergeyMoroz0703, line 201: `Could not retrieve applications properties`

app/selenium_ui/jira/pages/pages.py
```diff
@@ -6,6 +6,10 @@
 from selenium_ui.jira.pages.selectors import UrlManager, LoginPageLocators, DashboardLocators, PopupLocators, \
     IssueLocators, ProjectLocators, SearchLocators, BoardsListLocators, BoardLocators, LogoutLocators
 
+from util.data_preparation.jira_prepare_data import CLIENT, __check_rte_status
```
SergeyMoroz0703, line 9: Regarding Python naming conventions `__*` in the function name means that these functions are only for internal usage inside the class or module.

app/util/data_preparation/jira_prepare_data.py
mmizin: We can't import CLIENT from main().
Yes, I'll try to do it from jira_client

app/util/data_preparation/jira_prepare_data.py
mmizin: Already done )!

app/util/api/jira_clients.py
mmizin: Thx! Three times I've made changes and the last one forgot

app/util/api/jira_clients.py
```diff
@@ -195,3 +198,16 @@ def get_locale(self):
...
         return user_properties.json()['locale']
+
+    def get_applications_properties(self):
+        api_url = f'{self.host}/rest/api/2/application-properties'
+        user_properties = self.get(api_url, "Could not retrieve application properties")
+        return user_properties.json()
+
+    def check_rte_status(self):
+        app_prop = self.get_applications_properties()
+        rte = [i['value'] for i in app_prop if i['id'] == 'jira.rte.enabled'][0]
```
SergeyMoroz0703, line 209: `rte=app_prop.get('jira.rte.enabled')
return rte=='true'`

mmizin: Unfortunately, we can't use _get()_ method because of next dict scheme
` {'id': 'jira.rte.enabled',
  'key': 'jira.rte.enabled',
  'name': 'jira.rte.enabled',
  'type': 'boolean',
  'value': 'true'},`

PR #271, opened 2020-07-13
DCA-627:Refactor the getting webdriver_visible variable

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-627


app/selenium_ui/conftest.py
```diff
@@ -71,6 +71,7 @@ def __read_input_file(file_path):
 
 selenium_results_file = ENV_TAURUS_ARTIFACT_DIR / 'selenium.jtl'
 selenium_error_file = ENV_TAURUS_ARTIFACT_DIR / 'selenium.err'
+webdriver_visible = os.getenv('WEBDRIVER_VISIBLE', False)
```
ometelytsia, line 74: It feels like we do not need this env variable at all now, because we use .yml file for this configuration.
Let's remove it completely.

ometelytsia: Also, need to clean README files in docs folder from this env variable.

PR #277, opened 2020-07-20
NONE fix verify_response default arguments

SergeyMoroz0703:


app/util/api/abstract_clients.py
```diff
@@ -77,6 +77,8 @@ def put(self, url: str, error_msg: str, body: dict = None, params=None):
         return response
 
     def __verify_response(self, response: Response, error_msg: str, expected_status_codes: list = None):
```
ometelytsia, line 79: did you try:
`expected_status_codes: list = list())`

PR #280, opened 2020-07-20
bitbucket/DCA-655:Add perf_user to retrieve data to which he has access

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-655


mmizin:
The users that we are using to retrieve test data have admin permissions, so we don't need to check their permissions.

PR #281, opened 2020-07-21
DCA-660:Add check if jira_user has admin permission

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-660

### DIFF
--- a/app/util/data_preparation/jira_prepare_data.py
+++ b/app/util/data_preparation/jira_prepare_data.py
@@ -153,6 +153,14 @@ def __check_current_language(jira_api):
                          f'Please change your profile language to "English (United States) [Default]"')
 
 
+def __check_for_admin_permission(jira_api):
+    user_permissions = jira_api.get_user_permissions()
+    if not any((user_permissions['permissions']['ADMINISTER']['havePermission'],
+                user_permissions['permissions']['SYSTEM_ADMIN']['havePermission'])):
+        raise SystemExit(f"The '{JIRA_SETTINGS.admin_login}' user "
+                         f"does not has admin permission to perform actions.")
+
+
 def main():
     print("Started preparing data")
 
@@ -161,6 +169,7 @@ def main():
 
     client = JiraRestClient(url, JIRA_SETTINGS.admin_login, JIRA_SETTINGS.admin_password)
 
+    __check_for_admin_permission(client)
     __check_current_language(client)
     dataset = __create_data_set(client)
     write_test_data_to_files(dataset)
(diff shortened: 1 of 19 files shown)


app/util/data_preparation/jira_prepare_data.py
```diff
@@ -153,6 +153,14 @@ def __check_current_language(jira_api):
                          f'Please change your profile language to "English (United States) [Default]"')
 
 
+def __check_for_admin_permissions(jira_api):
+    user_permissions = jira_api.get_user_permissions()
+    if not any((user_permissions['permissions']['ADMINISTER']['havePermission'],
+                user_permissions['permissions']['SYSTEM_ADMIN']['havePermission'])):
+        raise SystemExit(f"The '{JIRA_SETTINGS.admin_login}' user "
```
ometelytsia, line 160: It makes sense to get the user name from jira_api, but not from Settings. Also, `have` instead of `has`.
`raise SystemExit(f"The '{jira_api.user}' user does not have admin permissions.")`

app/util/data_preparation/jira_prepare_data.py
```diff
@@ -153,6 +153,14 @@ def __check_current_language(jira_api):
                          f'Please change your profile language to "English (United States) [Default]"')
 
 
+def __check_for_admin_permissions(jira_api):
+    user_permissions = jira_api.get_user_permissions()
+    if not any((user_permissions['permissions']['ADMINISTER']['havePermission'],
```
ometelytsia, line 158: It would be more crear to use `or` instead of any()
```
    if not (user_permissions['permissions']['ADMINISTER']['havePermission'] or
            user_permissions['permissions']['SYSTEM_ADMIN']['havePermission']):
```

app/util/data_preparation/jira_prepare_data.py
```diff
@@ -153,6 +153,14 @@ def __check_current_language(jira_api):
...
+                user_permissions['permissions']['SYSTEM_ADMIN']['havePermission'])):
```
ometelytsia, line 159: Could you pls check if `user_permissions['permissions']['ADMINISTER_PROJECTS']` if sufficient permissions level for toolkit needs.

app/util/data_preparation/jira_prepare_data.py
mmizin: Yes it makes sense

app/util/data_preparation/jira_prepare_data.py
mmizin: no problem, let's make it more readable

app/util/data_preparation/jira_prepare_data.py
mmizin: unfortunately no

PR #282, opened 2020-07-21
DCA-661:Add check if confluence user has admin permissons

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-661


app/util/api/confluence_clients.py
```diff
@@ -156,6 +156,12 @@ def get_locale(self):
             print(f"Warning: Could not get user locale: {error}")
         return language
 
+    def get_groups_membership(self, login):
```
ometelytsia, line 159: `login` name for variable is confusing. 
Let's change to `username`.

PR #287, opened 2020-08-05
Dca 570 update confluence dataset version

SergeyMoroz0703:


app/util/bitbucket/populate_db.sh
```diff
@@ -47,13 +47,14 @@ if [[ ! "${SUPPORTED_BITBUCKET_VERSIONS[@]}" =~ "${BITBUCKET_VERSION}" ]]; then
   # Check if --force flag is passed into command
   if [[ "$1" == "--force" ]]; then
     # Check if passed Bitbucket version is in list of supported
-    if [[ "${SUPPORTED_BITBUCKET_VERSIONS[@]}" =~ "$2" ]]; then
+    if [[ " ${SUPPORTED_BITBUCKET_VERSIONS[@]} " =~ " ${2} " ]]; then
```
ometelytsia, line 50: Could you pls explain how this extra spaces works?

SergeyMoroz0703: They need to be sure that "empty" version could not be set as "JIRA_VERSION" to avoid broken s3 link to db.dump and attachments respectively. In previous --force implementation it worked.

PR #327, opened 2020-10-09
fix custom actions

SergeyMoroz0703:


app/selenium_ui/confluence/modules.py
```diff
@@ -8,9 +8,10 @@
 def setup_run_data(datasets):
     user = random.choice(datasets["users"])
     page = random.choice(datasets["pages"])
-    if len(datasets["custom_pages"]) > 0:
-        custom_page = random.choice(datasets["custom_pages"])
-        datasets['custom_page_id'] = custom_page[0]
+    if 'custom_pages' in datasets:
```
mmizin, line 11: How do you think, maybe it will be nice to make like in jira/modules "if CUSTOM_ISSUES in datasets:" -> if 'CUSTOM_PAGES' in datasets:" 
???

mmizin (approved):

app/selenium_ui/confluence/modules.py
SergeyMoroz0703: In modules.py for Jira we have all keys as variables. I used this approach for Confluence because in modules.py for Confluence we do not have any hardcoded dataset keys at all. 
I'll refactor current logic for modules.py for Confluence in this case.

astashys (approved):

PR #344, opened 2020-10-16
Reduced cache size for Docker

sstopkin:
Original image size:
dc-app-performance-toolkit             3.89GB
New image size:
dc-app-performance-toolkit-new     3.83GB

PR #346, opened 2020-10-16
Reduced cache size for Docker

sstopkin:
Original image size:
dc-app-performance-toolkit 3.89GB
New image size:
dc-app-performance-toolkit-new 3.83GB

Unfortunately Github doesn't allow to change origin for PR's created from web UI, so I recreated it using regular git. 
Fix of https://github.com/atlassian/dc-app-performance-toolkit/pull/344

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ RUN apt-get -y update \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1 \
   && python -m pip install --upgrade pip \
   && python -m pip install --upgrade setuptools \
-  && rm -rf /var/lib/apt/lists/*
+  && apt-get clean
 
 COPY requirements.txt /tmp/requirements.txt
 RUN pip install --no-cache-dir -r /tmp/requirements.txt
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,10 @@ RUN apt-get -y update \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1 \
   && python -m pip install --upgrade pip \
   && python -m pip install --upgrade setuptools \
-  && apt-get clean
+  && rm -rf /var/lib/apt/lists/*
 
 COPY requirements.txt /tmp/requirements.txt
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,13 +13,13 @@ RUN apt-get -y update \
   && update-alternatives --install /usr/bin/python python /usr/bin/python3.8 1 \
   && python -m pip install --upgrade pip \
   && python -m pip install --upgrade setuptools \
-  && apt-get clean
+  && rm -rf /var/lib/apt/lists/*
```
ometelytsia, line 16: I've tried only this line change with docker build --no-cache command and this does not reduce the size of the image, but this brokes possibility to install new packages with apt-get command in interactive mode.
I suggest removing this from the PR.

Dockerfile
```diff
@@ -13,13 +13,13 @@ RUN apt-get -y update \
...
 
 COPY requirements.txt /tmp/requirements.txt
-RUN pip install -r /tmp/requirements.txt
+RUN pip install --no-cache-dir -r /tmp/requirements.txt
```
ometelytsia, line 19: Good point.

Dockerfile
sstopkin: @ometelytsia if you will use apt in interactive mode after a few days (or month) since container build - apt cache will be outdated. So, in case if you would like to use apt - you should start with 'apt update' anyway.

ometelytsia: What is the benefit of this change if the image size stays the same? The size of this folder is 29Mb. Also, we do have a CI system that builds `dev` version of this container on a regular basis from the same Dockerfile and a couple of test containers that are based on the `atlassian/dcapt` container. So I do not see any benefits from this change, rather than additional troubles for our development pipeline.

sstopkin: @ometelytsia I don't see any potential issues in development process, but I don't want to force push any changes. So, I'm ok with keeping this 29 Mb, knowing about 3.8Gb of full container size :)

ometelytsia (approved):

SergeyMoroz0703 (approved):

PR #425, opened 2020-12-04
DCA-919:Update dependencies in requirements.txt file

mmizin:


requirements.txt
```diff
@@ -1,6 +1,6 @@
...
-pandas==1.1.2
-importlib-metadata==2.0.0
-zipp==2.2.0 # hard requirement of bzt 1.15.0
-bzt==1.15.0
-locust==1.2.3
\ No newline at end of file
+matplotlib==3.3.3
+pandas==1.1.4
+importlib-metadata==3.1.1
+zipp==3.4.0
```
ometelytsia, line 4: Why do we need this?

requirements.txt
```diff
@@ -1,6 +1,6 @@
...
+bzt==1.15.1
+locust==1.4.1
```
ometelytsia, line 6: We have to use the version of locust bundled with bzt 1.15.1, but not the latest from Pypi.

ometelytsia (changes requested):

requirements.txt
mmizin: I've run tests with locust(1.4.1) and everything works perfectly! Do you have a problem with this version?

ometelytsia: Bzt and locust compatibility is a bzt team responsibility, so for us locust is a bzt dependency.

requirements.txt
mmizin: Just have checked and ensured that we don't need it here

ometelytsia (approved):

PR #455, opened 2020-12-29
Update result summary and chart generation scripts

mmizin:


app/reports_generation/scripts/csv_aggregator.py
```diff
@@ -1,10 +1,18 @@
...
 from typing import List
-
 from scripts.utils import validate_str_is_not_blank, validate_file_exists, resolve_path
 
+from util.analytics.application_info import ApplicationSelector
+from util.analytics.analytics_utils import generate_test_actions_by_type
+
 RESULTS_CSV_FILE_NAME = "results.csv"
+APP_SPECIFIC_PREFIX = 'app_specific_'
+APP_NAME = 2
```
SergeyMoroz0703, line 14: Seems like magic numbers. Not sure that this is the best name for this vars

app/util/analytics/analytics_utils.py
```diff
@@ -6,8 +6,12 @@
 from datetime import datetime, timezone
 
 SUCCESS_TEST_RATE = 95.00
+SUCCESS_AVG_RT = 20
 OS = {'macOS': ['Darwin'], 'Windows': ['Windows'], 'Linux': ['Linux']}
 APP_SPECIFIC_TAG = 'APP-SPECIFIC'
+TEST_RATE = 0
+AVG_RT = 1
+EXCEPTIONS = ['jmeter_clone_repo_via_http', 'jmeter_clone_repo_via_ssh', 'selenium_create_pull_request']
```
SergeyMoroz0703, line 14: It would be great to add a comment there, why these actions in exception.

app/reports_generation/scripts/csv_aggregator.py
mmizin: I thought that it will be a magic number if use it in code without constants

PR #484, opened 2021-01-15
Jira nodes count fix

ometelytsia:


SergeyMoroz0703 (approved):

app/util/api/jira_clients.py
```diff
@@ -208,7 +208,7 @@ def get_nodes_count_via_rest(self):
         response = self.get(api_url, 'Could not get Jira nodes count', expected_status_codes=[200, 405])
         if response.status_code == 405 and 'This Jira instance is not clustered' in response.text:
             return 'Server'
-        nodes = [1 if node['state'] == "ACTIVE" else 0 for node in response.json()]
+        nodes = [1 if node['state'] == "ACTIVE" and node['alive'] else 0 for node in response.json()]
```
mmizin, line 211: What would you think about that variant
nodes = [node for node in response.json() if node['state'] == "ACTIVE" and node['alive']]
return len(nodes)

mmizin (approved):

PR #486, opened 2021-01-16
fix related to pull #327

totallytot:
Seems, the example in exstension_ui.py should be adjusted a bit too.


ometelytsia:
Hi @totallytot. Thank you for your input.
We made those changes to all products in this PR: https://github.com/atlassian/dc-app-performance-toolkit/pull/489

PR #508, opened 2021-02-01
added app specific label

SergeyMoroz0703:


app/reports_generation/scripts/chart_generator.py
```diff
@@ -1,11 +1,19 @@
...
 import matplotlib.pyplot as plt
 import pandas as pd
 from pandas import DataFrame
 
-from scripts.utils import validate_file_exists, validate_str_is_not_blank, validate_is_number
+from scripts.utils import validate_file_exists, validate_str_is_not_blank, validate_is_number, read_json_file
+
+APPS = ['jira', 'confluence', 'bitbucket', 'jsm']
+TEST_TYPES = ['selenium', 'jmeter', 'locust']
+DEFAULT_ACTIONS_PATH = '../util/default_test_actions.json'
```
ometelytsia, line 16: It feels like this should not be here at all.

app/reports_generation/scripts/chart_generator.py
```diff
@@ -22,10 +30,46 @@ def __resolve_and_expand_user_path(path: Path) -> Path:
     return path.resolve().expanduser()
 
 
+def __get_all_default_actions():
```
ometelytsia, line 33: This function is executed in a loop for every line, but read actions list should be a one-time operation.

app/reports_generation/scripts/chart_generator.py
```diff
@@ -22,10 +30,46 @@ def __resolve_and_expand_user_path(path: Path) -> Path:
...
+    full_actions_list = []
+    actions_data = read_json_file(DEFAULT_ACTIONS_PATH)
+    for app in APPS:
+        for test_type in TEST_TYPES:
+            for action in actions_data[app][test_type]:
+                full_actions_list.append(action)
+    return full_actions_list
+
+
+def __mark_as_app_specific(action_name: str):
```
ometelytsia, line 43: Confusion name and functionality of the method. Mark method should mark, but not decide what to mark.

app/reports_generation/scripts/chart_generator.py
```diff
@@ -22,10 +30,46 @@ def __resolve_and_expand_user_path(path: Path) -> Path:
...
+    if action_name not in __get_all_default_actions():
+        return f'\u2714{action_name}'
+    else:
+        return action_name
+
+
 def __read_file_as_data_frame(file_path: Path, index_col: str) -> DataFrame:
```
ometelytsia, line 50: Messed logic here.

app/reports_generation/scripts/chart_generator.py
astashys, line 16: Earlier we discussed choosing another approach.

PR #511, opened 2021-02-04
[JSM] DCA-1011: Selenium stability

mmizin:


app/selenium_ui/jsm/pages/agent_pages.py
```diff
@@ -159,4 +156,4 @@ def get_random_queue(self):
                                       ['All open', 'Recently resolved', 'Resolved past 7 days']
                                       and queue.text.partition('\n')[2] != '0'])
         random_queue.click()
-        self.wait_until_clickable(ViewQueueLocators.reporter)
+        self.wait_until_visible(ViewQueueLocators.queues_status, timeout=self.timeout)
```
ometelytsia, line 159: Why `timeout=self.timeout` is needed? `wait_until_visible` already has a default one.

mmizin: Because default timeout some times not enough to see loaded data from the queue.
test_1_selenium_agent_view_queues_medium-2021-02-02-14-20-47 299983

ometelytsia: Got it.

ometelytsia (approved):

app/selenium_ui/jsm/pages/agent_pages.py
SergeyMoroz0703: Better to fix timeouts in the base_page.py.
In the all wait_* functions need to use self.timeout (class variable) instead of TIMEOUT (global variable). 
Somewhere we already implement it, but I see that not in all "waiters".

mmizin: Fixed it

SergeyMoroz0703 (approved):

PR #582, opened 2021-04-20
[Confluence]:DCA:961 review jmeter actions on newer confluence versions

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-961


SergeyMoroz0703 (approved):

SergeyMoroz0703:
I strongly suggest to keep one naming convention for JMeter actions. After investigation, I found that some actions call `20 /rest/..`, some actions `20 rest/` and some of them like `20rest...`. So it might be a bit confusing.

ometelytsia (approved):

astashys (approved):

PR #583, opened 2021-04-20
[Confluence]:DCA-1104: Fix selenium login if node id doesn't exists

mmizin:
https://bulldog.internal.atlassian.com/browse/DCA-1104

### DIFF
--- a/app/selenium_ui/confluence/pages/pages.py
+++ b/app/selenium_ui/confluence/pages/pages.py
@@ -1,5 +1,4 @@
 from selenium_ui.base_page import BasePage
-from selenium.common.exceptions import NoSuchElementException
 
 from selenium_ui.confluence.pages.selectors import UrlManager, LoginPageLocators, AllUpdatesLocators, PopupLocators,\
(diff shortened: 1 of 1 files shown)


app/selenium_ui/confluence/pages/pages.py
```diff
@@ -39,8 +39,11 @@ def get_app_version(self):
         return text
 
     def get_node_id(self):
-        text = self.get_element(LoginPageLocators.footer_node_info).text
-        return text.split(':')[-1].replace(')', '').replace(' ', '')
+        try:
+            text = self.get_element(LoginPageLocators.footer_node_info).text
+            return text.split(':')[-1].replace(')', '').replace(' ', '')
+        except NoSuchElementException:
+            return 'UNDEFINED'
```
ometelytsia, line 46: Pls test following approach:

```
        if self.get_elements(LoginPageLocators.footer_node_info):
            text = self.get_element(LoginPageLocators.footer_node_info).text
            return text.split(':')[-1].replace(')', '').replace(' ', '')
        else:
            return "SERVER"
```

ometelytsia (approved):

PR #601, opened 2021-05-07
Fix taurus dockerfile version to avoid bintray

EugenMayer:
Since the old version still uses bintray as the `.deb` source, the build fails.
Using  `unstable` fixes it, since it is already fixed upstream at using https://github.com/Blazemeter/taurus/commit/7d939bb372c996f707be200744bab043c17b0fef#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557


ometelytsia:
Hi @EugenMayer. Thank you for raising this!
I've created PR https://github.com/atlassian/dc-app-performance-toolkit/pull/604/files to remove bintray repo manually for now (until the taurus PR you've mentioned is promoted to prod).
K6 is already installed in the latest prod image and dcapt does not need anything from bintray repo. So removing it from sources.list seems like a safe workaround.

Closing you PR for now.

PR #613, opened 2021-05-31
Crowd basic prepare data jmeter actions

SergeyMoroz0703:


app/crowd.yml
```diff
@@ -0,0 +1,86 @@
+---
```
ometelytsia, line 1: Need to clean up this file or create a separate tech debt ticket for this.
E.g. commented lines, correct crowd ports, default pass, etc.
Also, it makes sense to have another ticket for RPS approach?

app/util/data_preparation/crowd_prepare_data.py
```diff
@@ -0,0 +1,96 @@
...
+from util.api.crowd_clients import CrowdRestClient
+from util.project_paths import CROWD_USERS
+
+
+urllib3.disable_warnings(urllib3.exceptions.InsecureRequestWarning)
+
+
+USERS = "users"
+DEFAULT_USER_PASSWORD = 'password'
+DEFAULT_USER_PREFIX = 'performance'
```
ometelytsia, line 15: do we need _ here? e.g. 'performance_'.
Also, need to consider renaming to 'performance_crowd_' to avoid possible confusion with jira performance users.

app/util/data_preparation/crowd_prepare_data.py
```diff
@@ -0,0 +1,96 @@
...
+USER_SEARCH_CQL = f'name={DEFAULT_USER_PREFIX}*'
+ERROR_LIMIT = 10
+
+NECESSARY_USERS_COUNT = 10000
```
ometelytsia, line 19: What is that? Looks like this variable has to be renamed.

ometelytsia (approved):

app/crowd.yml
SergeyMoroz0703: Already created appropriate task

app/util/data_preparation/crowd_prepare_data.py
SergeyMoroz0703: Fixed

PR #638, opened 2021-06-17
DCR 166

nkvitko:
https://bulldog.internal.atlassian.com/browse/DCR-166


ometelytsia:
Declined in favor of https://github.com/atlassian/dc-app-performance-toolkit/pull/639

PR #664, opened 2021-08-20
Bamboo prepare_data and related API client

SergeyMoroz0703:


SergeyMoroz0703:
PEP8 will be fixed soon in this PR

app/bamboo.yml
```diff
@@ -0,0 +1,89 @@
+---
+settings:
+  artifacts-dir: results/bamboo/%Y-%m-%d_%H-%M-%S
+  aggregator: consolidator
+  verbose: false
+  env:
+    application_hostname: test-bamboo.atlassian.com    # Bamboo DC hostname without protocol and port e.g. test-bamboo.atlassian.com or localhost
+    application_protocol: http      # http or https
+    application_port: 8085            # 80, 443, 8080, 4990, etc
```
ometelytsia, line 9: let's remove 4990 from examples as this is a crowd port

app/bamboo.yml
```diff
@@ -0,0 +1,89 @@
...
+    secure: True                    # Set False to allow insecure connections, e.g. when using self-signed SSL certificate
+    application_postfix:            # e.g. /bamboo in case of url like http://localhost:4990/bamboo
```
ometelytsia, line 11: let's remove 4990 from examples as this is a crowd port

app/bamboo.yml
```diff
@@ -0,0 +1,89 @@
...
+    admin_login: admin
+    admin_password: admin
+    load_executor: jmeter            
+    concurrency: 1000               # number of concurrent threads to authenticate random users
```
ometelytsia, line 15: need to fix comment

app/util/api/bamboo_clients.py
```diff
@@ -0,0 +1,65 @@
+from util.api.abstract_clients import RestClient
+
+BATCH_SIZE_SEARCH = 500
+
+
+class BambooClient(RestClient):
+
+    def get_build_plans(self, start=0, max_result=100):
```
ometelytsia, line 8: It feels like the purpose of this method is mixed.
If this is an API method to get plans without pagination, then params `start` and `max_results` make sense.
If this is an API method to get N plans, then `start` param does not needed.

app/util/api/bamboo_clients.py
```diff
@@ -0,0 +1,65 @@
...
+
+            start += len(request.json()['searchResults'])
+
+        return content
+
+    def get_users(self, limit):
+        """
+        Retrieve a page of users. The authenticated user must have restricted
+        administrative permission or higher to use this resource.
+        :param start: The starting index of the returned boards. Base index: 0.
```
ometelytsia, line 37: the comment is off

app/util/api/bamboo_clients.py
```diff
@@ -0,0 +1,65 @@
...
+        :param limit: The maximum number of boards to return per page. Default: 50.
+        """
+        request = self.get(f'{self.host}/rest/api/latest/admin/users?limit={limit}',
```
ometelytsia, line 40: Does it have pagination?

app/util/api/bamboo_clients.py
```diff
@@ -0,0 +1,65 @@
...
+        :password: password.
+        :passwordConfirm: confirm password.
+        """
+        api_url = f'{self.host}/rest/api/latest/admin/users'
+        payload = {"name": username,
+                   "fullName": username,
+                   "email": f'{username}@example.com',
+                   "password": password,
+                   "passwordConfirm": password}
+        request = self.post(api_url, body = payload, error_msg="Could not create user")
```
ometelytsia, line 61: is this request returns something?

app/util/api/bamboo_clients.py
```diff
@@ -0,0 +1,65 @@
...
+        if request.ok:
```
ometelytsia, line 62: why do we need a check for `request.ok` if a post from the abstract client has built-in `__verify_response`?

app/util/api/bamboo_clients.py
SergeyMoroz0703: Nope, it returns all users set in `limit`

app/util/api/bamboo_clients.py
SergeyMoroz0703: Nope, there is no valuable response.

app/util/api/bamboo_clients.py
SergeyMoroz0703: Yap, sure, will fix it.

app/bamboo.yml
SergeyMoroz0703: Will remote port at all till the moment we will re-check the port from local installation.

PR #699, opened 2021-10-20
As when finished testing, with results accepted by Atlassian Marketplace

yarithphay-whs:


ometelytsia:
@yarithphay-whs https://stackoverflow.com/questions/18318097/delete-a-closed-pull-request-from-github

PR #708, opened 2021-10-29
Tech debt: improve get node info to work both on server and DC

opopovss:


ometelytsia (approved):

app/selenium_ui/jsm/pages/agent_pages.py
```diff
@@ -47,8 +47,14 @@ def get_app_version(self):
...
-        text = self.__get_footer_text()
-        return text.split(':')[-1].replace(')', '')
+        text = self.get_element(LoginPageLocators.footer).text
+        text_split = text.split(':')
+        if len(text_split) == 2:
+            return "SERVER"
+        elif len(text_split) == 3:
+            return text_split[2].replace(')', '')
+        else:
+            return f"Warning:  failed to get the node id from {text}"
```
OlehStefanyshyn, line 57: "Let's change text to f"Warning: failed to get the node information from '{text}'." - this way we could easily catch a case when text is an empty string."

Probably, you should change it here too

OlehStefanyshyn (approved):

PR #748, opened 2021-12-10
Docs/generator updates

ometelytsia:


README.md
```diff
@@ -146,7 +149,7 @@ pip install virtualenv
 ```
 2. Create new virtual env with python3:
 ```
-virtualenv venv -p python
+virtualenv venv -p full_path_to_python # e.g. use `where python3.9` to find the path
```
opopovss, line 152: Command where python3.9 -will not work, it's better to check where is all python installed where python and choose from there is 3.9 version path

PR #784, opened 2022-01-25
Added check of the selenium logout action

opopovss:
selenium_z_log_out     |   OK   | 100.00% |  1.110 - from 300 runs


app/selenium_ui/bamboo/pages/selectors.py
```diff
@@ -89,4 +89,5 @@ class JobConfigLocators:
 
 
 class LogoutLocators:
-    logout = (By.XPATH, "//a[@href='/userLogout.action']")
+    logout_url = UrlManager().logout_url()
+    login_button_link = (By.XPATH, "//a[@id='login']")
```
ometelytsia, line 93: why xpath if there is `id`?

ometelytsia (approved):

OlehStefanyshyn (approved):

PR #787, opened 2022-01-26
added check for user permission on db

j4ckr3d:
permission check by creation and drop of a TEST db, this way the main db used is still up and running without breaking jira by dropping it without having the possibility to re-create it.

### DIFF
--- a/app/util/jira/populate_db.sh
+++ b/app/util/jira/populate_db.sh
@@ -163,7 +163,7 @@ else
 fi
 echo "Current PostgreSQL version is $(psql -V)"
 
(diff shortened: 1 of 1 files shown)


ometelytsia:
Hi @j4ckr3d. Thank you for your contribution!
Could you pls re-create this PR against a `dev` branch?
Merges into `master` happen only during releases.

Thank you,
Alex

j4ckr3d:
Sorry, my mistake, i tought i already was on dev branch.
I'll move the check on dev and replicate it in other products you mentioned, thanks!

PR #804, opened 2022-02-15
conlfuence_clients.py modified

opopovss:
added "retry()" decorator  for "is_remote_api_enabled" check in confluence_clients.py

### DIFF
--- a/app/util/api/confluence_clients.py
+++ b/app/util/api/confluence_clients.py
@@ -124,6 +124,7 @@ class ConfluenceRestClient(RestClient):
 
         return search_results_list
 
+    @retry()
     def is_remote_api_enabled(self):
         api_url = f'{self.host}/rpc/xmlrpc'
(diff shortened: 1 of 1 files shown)


ometelytsia (approved):

app/util/api/confluence_clients.py
```diff
@@ -124,6 +124,7 @@ def search(self, cql, cqlcontext=None, expand=None, start=0, limit=500):
 
         return search_results_list
 
+    @retry()
```
ometelytsia, line 127: Pls double check the behavior when actually there is Remote API is disabled.
With retry, we still should report a correct error, but after 4 retries.

ometelytsia (approved):

app/util/api/confluence_clients.py
opopovss: Checked - an error appears:
`Exception: Confluence Remote API (XML-RPC & SOAP) is disabled. For further processing enable Remote API via General Configuration - Further Configuration - Remote API. Response code:[403]`

PR #810, opened 2022-03-30
Selenium 4 migrate

galynazholtkevych:


app/selenium_ui/base_page.py
```diff
@@ -57,57 +57,63 @@ def element_exists(self, selector):
...
 
-    def __wait_until(self, expected_condition, time_out=timeout):
+    def __wait_until(self, expected_condition, locator, time_out=timeout):
         message = f"Error in wait_until: "
         ec_type = type(expected_condition)
         if ec_type == AnyEc:
             conditions_text = ""
             for ecs in expected_condition.ecs:
-                conditions_text = conditions_text + " " + f"Condition: {str(ecs)} Locator: {ecs.locator}\n"
+                conditions_text = conditions_text + " " + f"Condition: {str(ecs)} Locator: {locator}\n"
```
ometelytsia, line 102: Could we double check how this works in case of AnyEC? We need to make sure that the correct locator is printed to error message.

PR #845, opened 2022-06-15
Add pytests to atlassians latest

fredrikattebrant:
As there were test failures when I tried to update our "2021 branch" with the latest from Atlassian (upstream) I now instead pick the pytests stuff for testing DM on to the Atlassian "HEAD".

Have launched testing but as they take a long time, they are not finished yet, thus kind of expecting having to push further updates. To be continue...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ outputs/
 *.pyc
 local_*.yml
 **jmeter.properties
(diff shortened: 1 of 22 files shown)


fredrikattebrant:
Sorry, should have gone to our fork. Bye bye.

PR #885, opened 2022-09-30
PERFT-100:  refactor browser metrics

Kryvetskyi:


app/util/confluence/browser_metrics.py
```diff
@@ -0,0 +1,14 @@
+
+metrics = {
```
SergeyMoroz0703, line 2: Please rename it to `browser_metrics`, to be precise.

app/selenium_ui/conftest.py
```diff
@@ -246,19 +246,22 @@ def get_requests_by_url(requests, url_path):
     return filtered_requests
 
 
-def get_wait_browser_metrics(webdriver):
+def get_wait_browser_metrics(webdriver, expected_metrics):
     attempts = 20
```
SergeyMoroz0703, line 250: Please reduce the attempts to 10, to have a total possible sleep time 5 sec. And need to retest it with 5 secs and 10 attempts.

app/selenium_ui/conftest.py
```diff
@@ -246,19 +246,22 @@ def get_requests_by_url(requests, url_path):
...
-    for i in range(0, attempts):
+    for i in range(attempts):
         requests = get_performance_logs(webdriver)
         requests_bulk = get_requests_by_url(requests, 'bulk')
-        for request_id, request in requests_bulk.items():
-            if 'browser.metrics.navigation' in str(request):
-                return requests_bulk
+        data.update(requests_bulk)
+
+        if all([metric in str(data) for metric in expected_metrics]):
```
SergeyMoroz0703, line 259: Maybe it is worth collecting the `keys` from the data and checking if the metric is in the keys? Need to test and discuss.

app/selenium_ui/conftest.py
```diff
@@ -246,19 +246,22 @@ def get_requests_by_url(requests, url_path):
...
+            return data
+
         print(f'Waiting for browser metrics, attempt {i}, sleep {sleep_time}')
         time.sleep(sleep_time)
+
```
SergeyMoroz0703, line 264: If metrics are not found need to print a WARNING message, like `WARNING: expected_metrics not found for the some_action`

app/selenium_ui/conftest.py
Kryvetskyi: Doesn't work properly with 10 attempts. Tested with 15 attempts, looks good to me.

SergeyMoroz0703 (approved):

PR #886, opened 2022-10-03
PERFT-71: Fix quick-view metrics

Kryvetskyi:


SergeyMoroz0703:
Need to create PR with the changes related to the current tasks, without previous changes.

PR #887, opened 2022-10-04
Perft 71: fix quick edit metrics

Kryvetskyi:


app/selenium_ui/confluence/modules.py
```diff
@@ -194,6 +194,7 @@ def measure():
 
         def sub_measure():
             page.go_to()
+            page.check_resources_loaded()
```
SergeyMoroz0703, line 197: `page.wait_for_resources_loaded()`

app/selenium_ui/confluence/pages/pages.py
```diff
@@ -83,6 +85,19 @@ def wait_for_comment_field(self):
     def click_edit(self):
         self.wait_until_clickable(PageLocators.edit_page_button).click()
 
+    def check_resources_loaded(self):
+        attempts = 3
+        for i in range(attempts):
+            time.sleep(1)
```
SergeyMoroz0703, line 91: sleep_time to variable

app/selenium_ui/confluence/pages/pages.py
```diff
@@ -83,6 +85,19 @@ def wait_for_comment_field(self):
     def click_edit(self):
         self.wait_until_clickable(PageLocators.edit_page_button).click()
 
+    def check_resources_loaded(self):
+        attempts = 3
```
SergeyMoroz0703, line 89: We do not need this variable. Need to break if True and continue iteration if not. Print message after full `for` loop.

SergeyMoroz0703 (approved):

app/selenium_ui/confluence/pages/pages.py
```diff
@@ -83,6 +85,20 @@ def wait_for_comment_field(self):
     def click_edit(self):
         self.wait_until_clickable(PageLocators.edit_page_button).click()
 
+    def wait_for_resources_loaded(self):
+        sleep_time = 0.1
```
ometelytsia, line 89: It would be nice to have timeout user-friendly:
```
timeout = 5 
sleep_time = 0.1
attempts = timeout // sleep_time
```

app/selenium_ui/confluence/pages/pages.py
```diff
@@ -83,6 +85,20 @@ def wait_for_comment_field(self):
...
+        attempts = 30
+        total_sleep = 0
+
+        for _ in range(attempts):
+            loaded = self.execute_js("return require('confluence-editor-loader/editor-loader').resourcesLoaded();")
+            if loaded:
+                return
+            time.sleep(sleep_time)
+            total_sleep += sleep_time
+            print(f'Total sleep time: {total_sleep}')
```
ometelytsia, line 99: The message is not very clear.
Imagine that you see a message in the log: 
`Total sleep time: 2.3`

app/selenium_ui/confluence/pages/pages.py
```diff
@@ -83,6 +85,20 @@ def wait_for_comment_field(self):
...
+        print('confluence-editor-loader resource could not be loaded')
```
ometelytsia, line 100: This error message is confusing. Also, do we want to continue test even if resource was not loaded?
Let's change to something like:
`print(f'WARNING: confluence-editor-loader resource was not loaded in {timeout} seconds.')`

app/selenium_ui/confluence/pages/pages.py
dmika1: I think timeout should be measured with real time - not attempts with sleep. Please modify the function to use start_time and `while time.time() - start_time < timeout`

app/selenium_ui/confluence/pages/pages.py
```diff
@@ -83,6 +85,20 @@ def wait_for_comment_field(self):
     def click_edit(self):
         self.wait_until_clickable(PageLocators.edit_page_button).click()
 
+    def wait_for_resources_loaded(self):
```
dmika1, line 88: timeout should be passed as a parameter with a default value

dmika1 (changes requested):

PR #918, opened 2022-11-10
Added js wait for Confluence page is fully loaded

SergeyMoroz0703:


app/selenium_ui/base_page.py
```diff
@@ -56,6 +57,20 @@ def element_exists(self, selector):
         by, locator = selector[0], selector[1]
         return True if self.driver.find_elements(by, locator) else False
 
+    def wait_for_js_statemant(self, key, value, timeout=timeout):
+        start_time = time.time()
+        print(f'Waiting for {key} is equal to {value}: {timeout} s.')
+        last_js_value = None
+        while time.time() - start_time < timeout:
+            js_current_value = self.execute_js(f'return {key}')
+            last_js_value = js_current_value
```
ometelytsia, line 66: not clear why do you need noth current and last value? It looks like only one value should be enough.

app/selenium_ui/base_page.py
```diff
@@ -56,6 +57,20 @@ def element_exists(self, selector):
...
+            if js_current_value == value:
+                print(f'{key} == {value} after {time.time() - start_time} s.')
+                break
+        else:
+            raise SystemExit(f'{key} is not equal to {value} for {timeout} s. Current {key} value is {last_js_value}')
```
ometelytsia, line 71: Do we really want to exit whole bzt execution in this case?

ometelytsia: pls test unhappy path when page was not loaded in timeout period

ometelytsia (approved):

opopovss (approved):

PR #932, opened 2022-11-17
Dca 1751 bump bzt version

dmika1:


ometelytsia:
@dmika1 I believe we need to bump bzt version in docker file as well

ometelytsia (approved):

opopovss (approved):

PR #970, opened 2023-01-17
Fix the regex to extract the pageId from the page source

nanux:
Confluence 8.1 introduced additional whitespace into the HTML tags; we need to change the regular expression extracting the `pageId` to account for it.

**Version 8.0**
```
<a  id="editPageLink" href="/wiki/pages/editpage.action?pageId=1004923276"
```

**Version 8.1**
```
<a  id="editPageLink"                href="/wiki/pages/editpage.action?pageId=1004923276"
```

### DIFF
--- a/app/jmeter/confluence.jmx
+++ b/app/jmeter/confluence.jmx
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.4.3">
+<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.5">
   <hashTree>
     <TestPlan guiclass="TestPlanGui" testclass="TestPlan" testname="upload_attachments" enabled="true">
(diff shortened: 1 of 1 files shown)


app/jmeter/confluence.jmx
```diff
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.4.3">
+<jmeterTestPlan version="1.2" properties="5.0" jmeter="5.5">
```
nanux, line 2: I have the 5.5 `jmeter`, I am not sure if this will be a problem. Please let me know if I should revert the change.

jjeongatl (approved):

app/jmeter/confluence.jmx
ometelytsia: Should not be a problem, but I would suggest keeping 5.4.3 here for consistency, as `confluence.yml` file has this version pinned.

ometelytsia (approved):

SergeyMoroz0703 (approved):

PR #1038, opened 2023-05-19
Tui tool

dmika1:


requirements.txt
```diff
@@ -8,4 +8,7 @@ selenium==4.9.0
 filelock==3.12.0
 packaging==23.1
 prettytable==3.7.0
+textual==0.23.0
```
ometelytsia, line 11: Could you pls add comments, to make it clear what these 3 dependency are related to TUI

PR #1041, opened 2023-05-26
DCA-2024: Improve jtl-to-csv.py script. Changed convert_to_csv func to only process results after first minute of the run

opopovss:


app/util/jtl_convertor/jtls-to-csv.py
```diff
@@ -133,29 +133,37 @@ def __pathname_pattern_expansion(args: List[str]) -> List[str]:
...
-        else:
-            # Get and update processed row with current label
-            processed_sample = [row for row in csv_list if row[LABEL] == jtl_sample['label']][0]
-            processed_sample[SAMPLES] = processed_sample[SAMPLES] + 1  # Count samples
-            processed_sample[ELAPSED_JTL_TMP].append(int(jtl_sample[ELAPSED_JTL]))  # list of elapsed values
-            processed_sample[SUCCESS_JTL_TMP].append(jtl_sample[SUCCESS_JTL].lower())  # list of success values
+        current_timestamp = int(jtl_sample['timeStamp'])
+
+        # Only process lines after 60 seconds (60000 milliseconds) from the first timestamp
+        if (current_timestamp - first_timestamp) >= 60000:
```
ometelytsia, line 151: need to move this value to configuration (in seconds)

dmika1 (approved):

app/reports_generation/scale_profile.yml
```diff
@@ -18,3 +18,4 @@ title: "DCAPT Scale Testing"
 image_height_px: 1000
 image_width_px: 1600
 judge: false
+cold_start: 60
```
ometelytsia, line 21: It would be nice to add a comment, something like:
# skip the first N seconds of the run results as the system is warming up

PR #1077, opened 2023-07-20
Update dcapt-small.tfvars Bitbucket 7.21.11 EBS snapshot id

ukuhnhardt:
Bitbucket 7.21.11 small EBS snapshot id is wrong (identical with BB 8.8.3)


ometelytsia:
New versions with correct snapshot ids would be released with next release.

PR #1082, opened 2023-07-26
[DO NOT MERGE] Bump chromedriver version up to 115

ometelytsia:
[DO NOT MERGE] Bzt is not able to download chromedriver 115 as they do not adapt yet new changes in chromedriver distribution.

### DIFF
--- a/app/bamboo.yml
+++ b/app/bamboo.yml
@@ -123,7 +123,7 @@ modules:
       httpsampler.ignore_failed_embedded_resources: "true"
   selenium:
     chromedriver:
-      version: "114.0.5735.90" # Supports Chrome version 114. You can refer to http://chromedriver.chromium.org/downloads
+      version: "115.0.5790.102" # Supports Chrome version 115. You can refer to https://googlechromelabs.github.io/chrome-for-testing/
 reporting:
 - data-source: sample-labels
   module: junit-xml
--- a/app/bitbucket.yml
+++ b/app/bitbucket.yml
@@ -90,7 +90,7 @@ modules:
       httpsampler.ignore_failed_embedded_resources: "true"
   selenium:
     chromedriver:
-      version: "114.0.5735.90" # Supports Chrome version 114. You can refer to http://chromedriver.chromium.org/downloads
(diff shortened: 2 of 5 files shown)


mims-github:
Hey @ometelytsia 
I jump in here, hope it's ok.

Problem:
We had problems with version mismatches from chrome installed in Dockerfile (stable (latest) => 116) and bzt chromedriver version (114) .
Suggestion:
Could you use in both places fixed versions, which are compatible with each other?
So e.g. we would suggest to use in https://github.com/atlassian/dc-app-performance-toolkit/blob/master/Dockerfile#L16 a link like: https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_114.0.5735.90-1_amd64.deb instead.

What do you think?

ometelytsia:
> Hey @ometelytsia I jump in here, hope it's ok.
> 
> Problem: We had problems with version mismatches from chrome installed in Dockerfile (stable (latest) => 116) and bzt chromedriver version (114) . Suggestion: Could you use in both places fixed versions, which are compatible with each other, so that it's not possible that there is a version mismatch if the upgrade rhythm is out of sync? So e.g. we would suggest to use in https://github.com/atlassian/dc-app-performance-toolkit/blob/master/Dockerfile#L16 a link like: https://dl.google.com/linux/chrome/deb/pool/main/g/google-chrome-stable/google-chrome-stable_114.0.5735.90-1_amd64.deb instead. Inspiration source: https://unix.stackexchange.com/questions/233185/install-older-versions-of-google-chrome-stable-on-ubuntu-14-10
> 
> What do you think?

Hi @mims-github, sorry somehow missed your comment.
The real problem was in the fact that chromedriver had changed new versions of distribution URLs.
So BZT tool needed some time to comply with the new changes: https://github.com/Blazemeter/taurus/issues/1753
Also, I've added ability into the Docker file to install any specific version of the chrome if needed:https://github.com/atlassian/dc-app-performance-toolkit/pull/1101/files

mims-github:
Thanks a lot. :clap:

PR #1222, opened 2024-01-08
Reporting/add specific and node count results

OlehStefanyshyn:


app/util/analytics/analytics_utils.py
```diff
@@ -154,6 +154,16 @@ def generate_report_summary(collector):
     write_to_file(pretty_report, summary_report_file)
 
 
+def rename_results_dir(collector):
```
ometelytsia, line 157: If relates to defining nodes only, so no sense to call os.rename twice.
Something like:
```
def rename_results_dir(collector):
    app_specific = ""
    if bool(collector.app_specific_rates):
        app_specific = "_app-specific"
    new_results_dir_name = f"{collector.log_dir}_{collector.nodes_count}-node{app_specific}"
    os.rename(f'{collector.log_dir}', f'{new_results_dir_name}_app-specific')
    print("Results directory was renamed. Old name: {collector.log_dir}}, new name: {new_results_dir_name}")
```

ometelytsia (approved):

PR #1238, opened 2024-01-15
fix resolution jira request issue

SergeyMoroz0703:


app/locustio/common_utils.py
```diff
@@ -125,6 +125,10 @@ def failure_check(self, response, action_name):
         if hasattr(response, 'error') or not response:
             if 'login' in action_name:
                 self.login_failed = True
+            try:
```
ometelytsia, line 128: Idea:
Verify if response is json type:
```
if response.headers.get('Content-Type') == 'application/json':
```

ometelytsia (approved):

PR #1333, opened 2024-03-11
Dca 2209 chromedriver check

SergeyMoroz0703:


app/util/conf.py
```diff
@@ -48,6 +48,7 @@ def __init__(self, config_yml):
         self.custom_dataset_query = self.get_property('custom_dataset_query') or ""
         self.verbose = self.settings['verbose']
         self.total_actions_per_hour = self.get_property('total_actions_per_hour')
+        self.chromedriver_version = self.obj['modules']['selenium']['chromedriver']['version']
```
ometelytsia, line 51: Why `self.chromedriver_version` is not in BaseAppSettings ?

app/util/pre_run/environment_compliance_check.py
```diff
@@ -60,6 +72,21 @@ def validate_application_config(processors, app_name_upper, app_settings, min_de
...
 
+def validate_chromedriver_version(app_name, app_settings):
+    options = webdriver.ChromeOptions()
+    options.add_argument("--headless")
+    driver = webdriver.Chrome(options=options)
+    current_chrome_version = version.parse(
+        driver.capabilities['browserVersion'])
+    current_chromedriver_version = version.parse(
+        app_settings.chromedriver_version)
+    if current_chromedriver_version != current_chrome_version:
```
ometelytsia, line 83: I suppose we need to compare only the major version, but not the full version.
E.g. My current latest Chrome is `122.0.6261.112`, but the latest chomedriver is `122.0.6261.111`

app/util/conf.py
SergeyMoroz0703: Its necessary to avoid failure on any crowd run.

ometelytsia: Initialize with default None?

app/util/pre_run/environment_compliance_check.py
SergeyMoroz0703: Fixed

ometelytsia (approved):

PR #1517, opened 2024-08-21
CONFSRVDEV-33309: Modify confluence.jmx to avoid unnecessary URL redirection when visiting a page or blog post

tamatlas:
When a page is visited via the legacy /display path or viewpage.action without the noRedirect query parameter, it will automatically redirect to a new URL in the form /spaces/<space-key>/pages/<page-id>/<page-title-slug>
In this PR, we add the noRedirect query parameter to the viewpage.action URL or change the /display path to the new format.

### DIFF
--- a/app/jmeter/confluence.jmx
+++ b/app/jmeter/confluence.jmx
@@ -67,7 +67,7 @@
         </collectionProp>
       </HeaderManager>
       <hashTree/>
-      <ThreadGroup guiclass="ThreadGroupGui" testclass="ThreadGroup" testname="Confluence" enabled="true">
+      <ThreadGroup guiclass="ThreadGroupGui" testclass="ThreadGroup" testname="Confluence">
         <intProp name="ThreadGroup.num_threads">1</intProp>
         <intProp name="ThreadGroup.ramp_time">0</intProp>
         <longProp name="ThreadGroup.duration">3600</longProp>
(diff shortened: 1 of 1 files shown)


app/jmeter/confluence.jmx
```diff
@@ -2347,8 +2361,8 @@ vars.put(&quot;page_title&quot;, &quot;jmeter_create_and_edit_page:create_page -
                 </JSR223PreProcessor>
                 <hashTree/>
               </hashTree>
-              <HTTPSamplerProxy guiclass="HttpTestSampleGui" testclass="HTTPSamplerProxy" testname="260 display/${space_key}/${page_title}" enabled="true">
-                <stringProp name="HTTPSampler.path">${application.postfix}/display/${space_key}/${page_title}</stringProp>
+              <HTTPSamplerProxy guiclass="HttpTestSampleGui" testclass="HTTPSamplerProxy" testname="260 /spaces/${space_key}/pages/${ajs-content-id}/${page_title}" enabled="true">
```
rjatkins, line 2364: Can we make this conditional on whether this is running vs Confluence 9.1+?

tamatlas: right. This URL won't work with older Confluence versions. Let me check if we can wrap a If Controller for this element

tamatlas: `If Controller` added to view the page based on the`confluence-version-number` variable. The variable is computed with the formula: major-number * 100 + minor-number (the patch number can be ignored).

app/jmeter/confluence.jmx
```diff
@@ -334,7 +334,11 @@
               <stringProp name="filename"></stringProp>
               <stringProp name="cacheKey">true</stringProp>
               <stringProp name="script">log.info(&quot;Logged in user - thread number: ${__threadNum}&quot;)
-log.info(&quot;Confluence version: ${confluence-version}&quot;)</stringProp>
+log.info(&quot;Confluence version: ${confluence-version}&quot;)
+String confluenceVersion = vars.get(&apos;confluence-version&apos;);
+String[] versionNumbers = confluenceVersion.split(&apos;[.]&apos;)
+vars.put(&quot;confluence-version-number&quot;, &quot;&quot; + (versionNumbers[0].toInteger() * 100 + versionNumbers[1].toInteger()))
```
ometelytsia, line 340: This calculation is not clear.
E.g. 9.0.2 would be translated to 900

ometelytsia: and 9.2.0 -> 902

ometelytsia (approved):

PR #1524, opened 2024-08-27
added wget .bzt dir

SergeyMoroz0703:


Dockerfile
```diff
@@ -29,7 +29,10 @@ RUN if [ "$CHROME_VERSION" = "latest" ]; then wget -O google-chrome.deb $CHROME_
 COPY requirements.txt /tmp/requirements.txt
 RUN pip install --no-cache-dir -r /tmp/requirements.txt
 
-RUN rm -rf /root/.bzt/jmeter-taurus/
+RUN wget https://blazemeter-tools.s3.us-east-2.amazonaws.com/bzt.tar.gz -O /tmp/bzt.tar.gz && \
```
ometelytsia, line 32: Need to think if we need any versioning for bzt.tar.gz

ometelytsia (approved):

PR #1798, opened 2025-08-13
DCA-2517 Update bamboo version to 10.2.7

ometelytsia:


app/util/k8s/dcapt.tfvars
```diff
@@ -399,8 +399,8 @@ crowd_additional_jvm_args = ["-Dupm.plugin.upload.enabled=true"]
 # By default, latest supported by DCAPT version is set.
 # https://hub.docker.com/r/atlassian/bamboo/tags
 # https://hub.docker.com/r/atlassian/bamboo-agent-base/tags
-bamboo_version_tag       = "10.2.2"
-bamboo_agent_version_tag = "10.2.2"
+bamboo_version_tag       = "10.2.7"
```
vkavalenka80, line 402: Isn't 11.0.2 is the latest? Or it's not supported by DCAPT?

vkavalenka80: Please disregard, I think I get it. This is because Renovate bamboo-spec bump and we have to do these versions tag to match. Correct?

PR #1828, opened 2025-08-27
Added error handling for terminating S3

bgospodarek:
Added error handling for terminating S3 clusters, added warning log if tags are missing for cluster.

### DIFF
--- a/app/util/k8s/terminate_cluster.py
+++ b/app/util/k8s/terminate_cluster.py
@@ -981,7 +981,11 @@ def delete_expired_tf_state_s3_buckets():
         for bucket in response["Buckets"]:
             if bucket_name_template in bucket["Name"]:
                 created_date = bucket["CreationDate"]
-                tags = s3_client.get_bucket_tagging(Bucket=bucket["Name"])["TagSet"]
+                try:
+                    tags = s3_client.get_bucket_tagging(Bucket=bucket["Name"])["TagSet"]
+                except s3_client.exceptions.NoSuchTagSet:
+                    logging.warning(f"S3 bucket {bucket['Name']} does not have any tags.")
+                    continue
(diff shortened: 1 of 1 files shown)


app/util/k8s/terminate_cluster.py
```diff
@@ -981,7 +981,11 @@ def delete_expired_tf_state_s3_buckets():
         for bucket in response["Buckets"]:
             if bucket_name_template in bucket["Name"]:
                 created_date = bucket["CreationDate"]
-                tags = s3_client.get_bucket_tagging(Bucket=bucket["Name"])["TagSet"]
+                try:
+                    tags = s3_client.get_bucket_tagging(Bucket=bucket["Name"])["TagSet"]
+                except s3_client.exceptions.NoSuchTagSet:
+                    logging.warning(f"S3 bucket {bucket['Name']} does not have any tags.")
+                    continue
```
ometelytsia, line 988: Let's make a re-raise exception here with a clear error message containing the bucket name.
It would be better to fail the plan and have an engineer to take a look and decide what to do with a bucket, rather just skipping.

app/util/k8s/terminate_cluster.py
```diff
@@ -981,7 +981,11 @@ def delete_expired_tf_state_s3_buckets():
         for bucket in response["Buckets"]:
             if bucket_name_template in bucket["Name"]:
                 created_date = bucket["CreationDate"]
-                tags = s3_client.get_bucket_tagging(Bucket=bucket["Name"])["TagSet"]
+                try:
+                    tags = s3_client.get_bucket_tagging(Bucket=bucket["Name"])["TagSet"]
+                except s3_client.exceptions.NoSuchTagSet:
```
vkavalenka80, line 986: Wondering if it throws this exception... Logs say that it is botocore.exceptions.ClientError . Is it being caught here?

ometelytsia (approved):

PR #1999, opened 2026-01-22
Remove Python 3.9 support and bump requirements

bgospodarek:
* Updated requirements.txt
* Removed support for Python 3.9
* Added disclaimer in readme.md for locust user 

---
 <strong>You don't have access to Rovo Dev Standard</strong>
Ask your organization admin to add you to Rovo Dev Standard.
[More about code review errors](https://support.atlassian.com/rovo/docs/troubleshoot-rovo-dev-code-reviews/)


vkavalenka80 (commented):
This does not work for me. I created new venv using Pyrhon3.9 and tried to install requirements. They failed saying 
`ERROR: Cannot install -r requirements.txt (line 9) and urwid==2.1.2 because these package versions have conflicting dependencies.
                                                                                                                                                                                                                                                                                                                              
The conflict is caused by:
    The user requested urwid==2.1.2
    bzt 1.16.45 depends on urwid==2.6.16
`

README.md
```diff
@@ -75,6 +75,8 @@ source venv/bin/activate
 4. Install dependencies:
 ```
 pip install -r requirements.txt
+
+pip install urwid==2.1.2 --force-reinstall # if you intend to use locust as main load-executor while runnning tests
```
vkavalenka80: I'd suggest to move that to
5. Optional: if only you intend to use locust as main load-executor while running tests

requirements.txt
```diff
@@ -1,13 +1,13 @@
-matplotlib==3.10.7
+matplotlib==3.10.8
 pandas==2.3.3
-pytest==8.4.2
-locust==2.42.1
+pytest==9.0.2
+locust==2.43.1
 selenium==4.39.0
```
vkavalenka80, line 5: Please also update .yml files - I think we also have selenium version for some reason (though not sure why)

README.md
```diff
@@ -28,7 +28,7 @@ In case of technical questions, issues or problems with DC Apps Performance Tool
 ## Installation and set up
 
 #### Dependencies
-* Python 3.9 - 3.13 and pip
+* Python 3.10 - 3.14 and pip
```
vkavalenka80: I don't think we are 3.14 ready... I tried to run bzt locally and saw a few problems:
- a few regexp related warning from bzt at start-up - saw that one time but later disappeared
- selenium tests are not executed reporting "filelock" errors (???)
- downgrading urwid doesn't seem to help - but you may want to double check

Could you check that on your end?

bgospodarek: I've tested both bzt locally and in Docker container. Both worked without issues with Python 3.14.
(clean venv installation)


As for urwid, I've tested it for each version of Python, clean venv installation (one run with load executor set to locust and one with jmeter), did not get any traceback, errors, nor NotImplementedError.



I will try to run those tests once again then

vkavalenka80 (approved):

